PR #48, opened 2015-03-01
Added basic controller support

czipperz:
Adds implementations of the inputs package from lwjgl2 and LWJGLException, as well as adding jinput library.


Spasi:
LWJGL 3 already has controller support, via [GLFW](http://www.glfw.org/docs/latest/input.html#joystick). Besides, I don't think wrapping JInput is useful in any way. If you need something from JInput, you can use it directly.

PR #49, opened 2015-03-02
LWJGL ready for OSGi

kKdH:
Modified the build to create an additional artifact and fixed native library lookup to run lwjgl inside an osgi-environment.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -321,8 +321,12 @@
 	</target>
 
 	<target name="release" description="" depends="compile,javadoc">
-		<!-- Bundle classes -->
+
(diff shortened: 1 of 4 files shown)


Spasi:
The native library lookup has been fixed with 2543603d062a268bea6adf871c8afa592bddd97b.

As for the rest, there are no plans to support OSGi. If someone really needs OSGi compatibility, LWJGL releases are not signed; they can manually be rebundled with the necessary metadata.

io7m:
Apologies for digging up a corpse, but I've recently put together a small project that produces an OSGi bundle in an automated manner:

https://github.com/io7m/org.lwjgl

It appears to work, but hasn't had anywhere near enough testing yet. I'll be pushing the bundles to Central when the various platforms have been verified to work.

PR #56, opened 2015-04-15
Added an exclusion for reference counting on "NVIDIA CUDA" platform.

dhild:
Noticed that this test was failing when I ran it.
Incidentally, the run with "AMD Accelerated Parallel Processing" is working for me.


Spasi:
Hey Ryan,

Thanks for the PR, I've verified that it indeed fails on the Nvidia implementation in the same way it does on Intel's. Both AMD and Apple implementations work as expected, according to the OpenCL spec.

The test itself isn't very interesting though, so I simply removed the reference count checks.

PR #57, opened 2015-04-28
Added comments to OpenAl, OpenCL and OpenGL exceptions

Blunderchips:
Added comments to OpenAl, OpenCL and OpenGL exceptions


Spasi:
Thank you!

PR #58, opened 2015-05-01
Added some package-info documentation

Blunderchips:
Added some package-info documentation.


Spasi:
Hey Matthew,

The package statements are wrong; core is not part of the Java package hierarchy, it's the module name. All statements much be changed to "package org.lwjgl...".

Also, I don't think GLFW stands for  OpenGL FrameWork. Even if it does, the official name seems to be just GLFW.

Please update the PR and I will merge it.

Blunderchips:
will do

Spasi:
Thanks, merged manually.

PR #69, opened 2015-06-21
Make LWJGL 3 Java 1.5 compatible

httpdigest:
Since LWJGL 3 does not make use of any important Java 1.6 API or JNI functionality, we can just lower the bar on the required Java version even further to 1.5.
And who knows whoever is still using a Java 1.5 JRE because they cannot/don't want to update.
Oh and LWJGL 2 also looks like its 1.5 compatible. So people upgrading from 2 to 3 would have an easy life. :)

The only thing that needed adaptation is the change in the semantics of `@Override` between Java 1.5 and Java 1.6. So, with Java 1.5 source level we would not have a check for unimplemented interface methods in abstract classes anymore with this annotation. But since most code is generated anyway, this should be a minor point.

However, LWJGL 3 still needs to be built with at least a 1.6 JDK, since Kotlin requires that. But I guess the LWJGL 3 build infrastructure does that anyway, if not already with JDK1.7.

I successfully tested a build of the proposed changes under Windows 7 x64 with a VS2013 native build under jre1.5.0_22.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -132,40 +132,40 @@ License terms: http://lwjgl.org/license.php
 			<not>
 				<or>
 					<isset property="env.JAVA_HOME"/>
-					<isset property="env.JAVA6_HOME"/>
+					<isset property="env.JAVA5_HOME"/>
 				</or>
 			</not>
 		</condition>
 		<fail if="java.home.missing" message="Please set the JAVA_HOME environment variable."/>
 		<!--
-			JAVA6_HOME will be used for the bootclasspath. This will cause compilation to fail
-			if a Java 7+ API is used by mistake. Do not fail if it is not set, try to use
+			JAVA5_HOME will be used for the bootclasspath. This will cause compilation to fail
+			if a Java 6+ API is used by mistake. Do not fail if it is not set, try to use
 			JAVA_HOME instead and issue a warning if that's not a Java 6 SDK.
 		-->
 		<sequential>
 		<script language="javascript">
-		if ( !java.lang.System.getenv().containsKey("JAVA6_HOME") ) {
+		if ( !java.lang.System.getenv().containsKey("JAVA5_HOME") ) {
 			var version = java.lang.System.getProperty("java.specification.version");
-			if ( !version.equals("1.6") )
-				print("The environment variable JAVA6_HOME has not been set. Using JAVA_HOME (version " + version + ").");
+			if ( !version.equals("1.5") )
+				print("The environment variable JAVA5_HOME has not been set. Using JAVA_HOME (version " + version + ").");
 		}
 		</script>
 		</sequential>
-		<property name="env.JAVA6_HOME" value="${env.JAVA_HOME}" unless:set="env.JAVA6_HOME"/>
+		<property name="env.JAVA5_HOME" value="${env.JAVA_HOME}" unless:set="env.JAVA5_HOME"/>
 
 		<path id="jdk.boot.classpath.jars">
-			<fileset dir="${env.JAVA6_HOME}/jre/lib">
+			<fileset dir="${env.JAVA5_HOME}/jre/lib">
 				<include name="*.jar"/>
 			</fileset>
-			<fileset file="${env.JAVA6_HOME}/lib/tools.jar"/>
+			<fileset file="${env.JAVA5_HOME}/lib/tools.jar"/>
 		</path>
 		<path id="jdk.boot.classpath.applejars">
-			<fileset dir="${env.JAVA6_HOME}/bundle/Classes">
+			<fileset dir="${env.JAVA5_HOME}/bundle/Classes">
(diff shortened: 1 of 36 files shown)


Spasi:
The decision to make LWJGL 3 compatible with Java 6+ was made very early on. Java 5 was obsolete 3 years ago and remains so; it is also completely unsupported and unmaintained. The same could be said about Java 6 actually (e.g. on OS X), but at least there are still some recently updated builds out there (from both Oracle and Zing).

Even Java 7 reached the end of public updates threshold on April 2015. It's more likely to bump the minimum required Java version to 7/8 for the official 3.0 release, than go back to 5.

LWJGL 3 does not have existing deployments to support (it is not backwards compatible with LWJGL 2) and it's always a bad idea to rely on public JREs on user machines.

httpdigest:
> Java 5 was obsolete 3 years ago and remains so
Yes, of course. You are right. But that does not stop people from using it. And I am thinking about very old corporate computers for which there are still in-house projects to build and support.

> The decision to make LWJGL 3 compatible with Java 6+
But is there any reasoning behind this decision? I mean if there is no need for a library to be Java 1.x compatible, because it does not make use of API or language features, then, why make it so?
In my opinion the lowest requirement for a broadly used library is always the best choice.
But of course, if the decision is final, this issue can be closed.

> It's more likely to bump the minimum required Java version to 7/8 for the official 3.0 release
That would be most unfortunate in my opinion. But of course you decide.

Spasi:
> Yes, of course. You are right. But that does not stop people from using it. And I am thinking about very old corporate computers for which there are still in-house projects to build and support.

As I said, no one should rely on existing JRE installations. Most LWJGL users deploy their applications with private bundled JREs (as do most Java apps these days).

> But is there any reasoning behind this decision? I mean if there is no need for a library to be Java 1.x compatible, because it does not make use of API or language features, then, why make it so?
In my opinion the lowest requirement for a broadly used library is always the best choice.
But of course, if the decision is final, this issue can be closed.

I explained the reasons above. Technically, we could make LWJGL 3 compatible with Java 1.4 and up, but what would be the point? It only makes developing the library more painful, for no real benefit.

> That would be most unfortunate in my opinion. But of course you decide.

That was rhetorical; there would be something to gain from going to 7/8. Java 5 compatibility is pointless. But, as I said, the decision has been made, LWJGL 3 is Java 6+ compatible. The next logical step would be making LWJGL 4 require Java 8, for lambda callbacks. Unless Project Panama is available by then, in which case we'll have entirely different things to talk about.

httpdigest:
Don't get me wrong, all I am saying is that no writer of a framework can imagine all scenarios and environments that his framework is going to be used in.
And that he should always target for the least restrictions, unless absolutely unavoidable, so as to not patronize a client to use a specific minimum JRE version when technically unnecessary.
So if LWJGL 3.0 is going to absolutely have to make use of Java 7 or even Java 8 API and language features, then it be so.

Spasi:
The Java 7 language features are mostly syntactic sugar that we can live without. Try-with-resources is very useful, but on the user side. We also have a decent workaround for callbacks using Java 8 lambdas. API-wise, the only time I found myself missing stuff was when writing the Generator, but that's not part of the library.

httpdigest:
Yes, that's what I am thinking too. There seems to be very little gain in using Java 7 and 8 in a framework compared to a very high compatibility price to pay, just because "one was feature-happy or lazy." ;)
I know that non-capturing Lambdas in Java 8 can have a nice performance gain compared to handwritten anonymous classes which get instantiated with every call, but then again a framework should not make use of anonymous classes anyway. :)
But apart from that, do you have any nice new features for LWJGL 4? Apart from a modularized build?
I mean, considering the current state, LWJGL 3 is pretty perfect for me in its use. I cannot think of anything I would like to have changed on it.

Spasi:
I imagine that LWJGL 4 will be an opportunity to break source compatibility with 3 (just like 3 did with 2). This is obviously years down the line. I'm also very happy with LWJGL 3 and at the moment I can't think of anything other than Project Panama that would justify going to 4. I imagine that we'll get some kind of abstraction for pointers (possibly value-typed) and also interface-based structs. It would be a great opportunity to make the LWJGL API type-safe with respect to pointers.

Btw, there has been progress on Project Panama for quite some time now and it's starting to [shape up](http://hg.openjdk.java.net/panama/panama/jdk/rev/e2648e33b231).

httpdigest:
Panama looks very promising. Would be great if LWJGL can make use of it to get rid of the JNI overhead.
Okay, I am closing this PR now, because Java 1.6 is fine and 1.5 compatibility can as of now be obtain easily in a fork if desired.

Spasi:
Just noticed that the latest version of TestNG (the testing framework used by LWJGL) has dropped Java 6 support.

PR #70, opened 2015-06-24
Simple Gradle build file to deploy to SonaType or local Maven repo

badlogic:
Builds on top of the Ant `release` task. Any Gradle 2.x version will do. I didn't want to add the Gradle wrapper to the repo as it dumps a ton of files into the root. It's a quicky, so i'll likely clean it up. E.g. the build file currently pollutes the `release` directory.

### Generated artifacts
The artifacts that are generated are compatible with the old 2.x artifacts already in Central.

```
<dependency>
   <groupId>org.lwjgl</groupId>
   <artifactId>lwjgl</artifactId>
   <version>3.0.0a</version>
</dependency>
```

The natives are cleanly separated into separate artifacts as well, just like before:

```
<dependency>
   <groupId>org.lwjgl</groupId>
   <artifactId>lwjgl-platform</artifactId>
   <version>3.0.0a</version>
   <classifier>natives-windows</classifier>
</dependency>
```

```
<dependency>
   <groupId>org.lwjgl</groupId>
   <artifactId>lwjgl-platform</artifactId>
   <version>3.0.0a</version>
   <classifier>natives-linux</classifier>
</dependency>
```

```
<dependency>
   <groupId>org.lwjgl</groupId>
   <artifactId>lwjgl-platform</artifactId>
   <version>3.0.0a</version>
   <classifier>natives-osx</classifier>
</dependency>
```

### Deploying to the local repo
```
ant release
gradle uploadArchives
```
Puts artifacts in the local Maven repository.

### Deploying to SonaType
Since everything Maven is terrible, you'll need to do some setup first. [Create a SonaType account](http://central.sonatype.org/pages/ossrh-guide.html). You should already have one. Note down the credentials (user/password).

The version you publish is defined in line 5 of `build.gradle` (`project.version`). The `-SNAPSHOT` suffix will be automatically attached if you release snapshots.

#### Publishing Snapshots
```
ant release
gradle -Psnapshot -PsonatypeUsername=<yourusername> -PsonatypePassword=<yourpassword> uploadArchives
```

After that, your users can use snapshot builds by adding the SonaType snapshot repository to their Maven/Gradle.

**Gradle**
```
repositories {
    mavenCentral()
    mavenLocal()
    maven {
        url "https://oss.sonatype.org/content/repositories/snapshots"
    }
}
```

**Maven**
```
<repository>
         <id>snapshots-repo</id>
         <url>https://oss.sonatype.org/content/repositories/snapshots</url>
         <releases><enabled>false</enabled></releases>
         <snapshots><enabled>true</enabled></snapshots>
</repository>
```

#### Publishing Releases
You'll need a GPG key pair, and also register it with some keyserver. [Here's a guide](http://blog.sonatype.com/2010/01/how-to-generate-pgp-signatures-with-maven/#.VYs14hMrLMU). You should already have this for your 2.x Maven Central releases.

First, tag your last commit in Git. Next, make sure the `project.version` is set properly in the `build.gradle`file. Then do this:

```
ant release
gradlew -Prelease -PsonatypeUsername=<username> -PsonatypePassword=<password> -Psigning.keyId=<key-id> -Psigning.password=<key-password> -Psigning.secretKeyRingFile=<path/to/your/keyring.gpg uploadArchives
```

This will sign and publish your release artifacts to SonaType. You then have to login to https://oss.sonatype.org/#welcome and close and release the thing. Go to Staging Repositories, scroll down, select your thing, click Close. Wait until its closed (refresh), then select the thingy again and press release. Your artifacts will then be synched with Maven Central.

If something went wrong before you synched to Maven Central, fear not. Just drop the thing in SonaType and start from scratch.

Once everything is published, increase the `project.version` in the `build.gradle` to your next release version.

I'm happy to do the releases for you if you want, but that would require the sonatype credentials and the gpg key.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,5 @@
 *.wav
 *.zip
 touch.txt
+/.gradle
+/build
--- a/build.gradle
+++ b/build.gradle
@@ -1,0 +1,125 @@
+apply plugin: "maven"
+apply plugin: "signing"
+
+project.group = "org.lwjgl"
+project.version = "3.0.0a"
+
+def isDevBuild
+def isCiBuild
+def isReleaseBuild
+
+def repositoryUrl
+def repositoryUser
+def repositoryPassword
+
+//set build variables based on build type (release, continuous integration, development)
+if(hasProperty("release")) {
+    isReleaseBuild = true
+    repositoryUrl = "https://oss.sonatype.org/service/local/staging/deploy/maven2/"                             
+    repositoryUser = sonatypeUsername
+	repositoryPassword = sonatypePassword
+    println "Performing release build"
+} else if (hasProperty("snapshot")) {
+    isCiBuild = true
+    version += "-SNAPSHOT"
+    repositoryUrl = "https://oss.sonatype.org/content/repositories/snapshots/"
+    repositoryUser = sonatypeUsername
+	repositoryPassword = sonatypePassword
+    println "Performing snapshot build"
+} else {
+    isDevBuild = true
+    repositoryUrl = repositories.mavenLocal().url
+    repositoryUser = ""
+	repositoryPassword = ""
+    println "Performing local build"
+}
+
+artifacts {
+  archives file: file("release/jar/lwjgl.jar"), name: "lwjgl", type: "jar"
+  archives file: file("release/src.jar"), name: "lwjgl", type: "jar", classifier: "sources"
+  archives file: file("release/jar/lwjgl-natives-windows.jar"), name: "lwjgl-windows", type: "jar", classifier: "natives-windows"
+  archives file: file("release/jar/lwjgl-natives-macosx.jar"), name: "lwjgl-macosx", type: "jar", classifier: "natives-osx"
+  archives file: file("release/jar/lwjgl-natives-linux.jar"), name: "lwjgl-linux", type: "jar", classifier: "natives-linux"
+}
+
+if(isReleaseBuild) {
+    signing {
+        sign configurations.archives
+    }
+} else {
+    task signArchives {
+        // do nothing
+    }
+}
+
+Closure defaultPom = {
+	project {
+		packaging "jar"        
+		description 'LWJGL'
+		url 'http://www.lwjgl.org'
+
+		scm {
+		   url 'scm:git@github.com:LWJGL/lwjgl3.git'
+		   connection 'scm:git@github.com:LWJGL/lwjgl3.git'
+		   developerConnection 'scm:git@github.com:LWJGL/lwjgl3.git'
+		}
+
+		licenses {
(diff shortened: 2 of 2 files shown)


kenzierocks:
You may want to retarget this for the `gradle` branch, just a thought.

badlogic:
Honestly, i think you should stick with the Ant build. Gradle is a major pain in the butt, and i see no benefit in using Gradle for the general build. This is non-invasive and self-contained, and almost easy to understand.

Give up on the Gradle build and focus on the code instead.

badlogic:
Eh, nm, just saw the Gradle branch. Not much Gradle in there yet. But i guess it has the new modules separation as well. Let me retarget.

badlogic:
Actually, to much work, it's 1:25am over here. I let you decide what to do with the single `build.gradle` file :)

Spasi:
Thanks Mario, for the gradle script and the deployment tutorial. I don't have access to the org.lwjgl group atm, it will take some time to resolve the situation.

Don't worry about the gradle branch, I haven't pushed the latest changes anyway. I'll add your script manually when it's ready.

I'm wondering why you think gradle is a pita. I'm starting to form a similar opinion, but it would be interesting to hear from someone with more experience. My top complains would be:

- Multiple ways to do the same thing. No two scripts are alike, which I find really strange for a build system that favors convention over configuration.
- Performance is a joke, even with the daemon. Seriously, multiple seconds for stuff that are instant with Ant.

On the other hand, it certainly looks like that the Gradle scripts will be cleaner and more maintainable than the Ant scripts.

badlogic:
In short: it's a mess, as big if not bigger than Ant. Wgile the scripts may
look more maintainable at first, they'll only be maintainable by yourself,
especially if you do anything other than standard Java projects. The LOC
count is likely the same as with Ant. You also get a metric ton of magic
nobody understands and Gradle is still a moving target. I'd recommend
sticking with Ant and keep my simple Gradle file to deploy Maven artifacts
locally or to SonaType/Central.

b23prodtm:
I find this as good guidelines to Maven2 publishing. I'm sure it can help. 

Nervertheless I'm not used to the POM form for the desktop application.. In my opinion, POM is a pattern designed to maintain over Web Apps, which stand onto web HTML/JS API bindings (and the maven central dependencies), that update faster than OS SDK.

Regards

badlogic:
I fixed up the build and published 3.0.0a to Maven Central. I'm setting up a Jenkins job that publishes SNAPSHOT builds to SonaType if that's OK with you. For that, i had to increase the version in `build.gradle` to 3.0.0b. For the next release, we need to set the version to whatever you deem good.

badlogic:
Closed in favor of #71

PR #71, opened 2015-06-27
Gradle build file for Maven Central/SonaType deployment

badlogic:
Reworked and squashed commits.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,5 @@
 *.wav
 *.zip
 touch.txt
+.gradle/
+build/
--- a/build.gradle
+++ b/build.gradle
@@ -1,0 +1,156 @@
+apply plugin: "maven"
+apply plugin: "signing"
+
+project.group = "org.lwjgl"
+project.version = "3.0.0b"
+
+def isDevBuild
+def isCiBuild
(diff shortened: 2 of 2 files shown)


Spasi:
Hey Mario,

I had some time to test this script and the tutorial in #70, it works great and actually was much easier than I expected. I will merge this PR soon and will also make the build server publish snapshots to SonaType.

One question: why is `build.gradle` in the [artifact list](https://github.com/LWJGL/lwjgl3/pull/71/files#diff-c197962302397baf3a4cc36463dce5eaR41)?

badlogic:
I needed something to package for that specific artifact because Gradle
sucks. The build.gradle was the smallest file, so i eventually gave in and
used that.

Thanks for the merge! Did you setup a build job for snapshots? Still happy
to do that on my end if you want.

Spasi:
Yes, the same job that publishes nightly builds on the lwjgl website, now also uploads a snapshot to Sonatype.

> I needed something to package for that specific artifact because Gradle
sucks. The build.gradle was the smallest file, so i eventually gave in and
used that.

I'm not sure what you mean here. I have removed it and it seems to be working fine.

PR #76, opened 2015-08-07
Rename org.lwjgl.openal.Util to org.lwjgl.openal.ALUtil

SilverTiger:
For the sake of consistency I would propose to rename org.lwjgl.openal.Util to org.lwjgl.openal.ALUtil, but it may break existing code that is using the org.lwjgl.openal.Util class.


Spasi:
Yes, thank you.

PR #90, opened 2015-09-25
Add Eclipse IDE project files

httpdigest:
These include the .project and .classpath files needed for Eclipse Java projects with a minimum working configuration.

### DIFF
--- a/config/ide/eclipse/.classpath
+++ b/config/ide/eclipse/.classpath
@@ -1,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<classpath>
+	<classpathentry excluding="org/lwjgl/egl/|org/lwjgl/opengles/|org/lwjgl/glfw/GLFWNativeEGL.java" kind="src" path="modules/core/src/generated/java"/>
+	<classpathentry excluding="org/lwjgl/egl/|org/lwjgl/opengles/" kind="src" path="modules/core/src/main/java">
+		<attributes>
+			<attribute name="org.eclipse.jdt.launching.CLASSPATH_ATTR_LIBRARY_PATH_ENTRY" value="lwjgl3/libs"/>
+		</attributes>
+	</classpathentry>
+	<classpathentry excluding="org/lwjgl/demo/ovr/" kind="src" path="modules/core/src/test/java"/>
+	<classpathentry excluding="org/lwjgl/system/|org/lwjgl/PointerBuffer.java" kind="src" path="modules/templates/src/main/java"/>
+	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER"/>
+	<classpathentry kind="lib" path="libs/testng.jar"/>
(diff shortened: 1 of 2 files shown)


httpdigest:
Usage hint: Remember to copy those two files from `config/ide/eclipse` to the root folder of the lwjgl3 sources for Eclipse to work properly, and then import the project into Eclipse.

Spasi:
> Remember to copy those two files from config/ide/eclipse to the root folder of the lwjgl3 sources for Eclipse to work properly, and then import the project into Eclipse.

Is it possible to make it so this isn't required? With IntelliJ you can have the project files in any directory and all paths will be configured relative to it. See [this module](https://github.com/LWJGL/lwjgl3/blob/master/config/ide/idea/Core.iml) for example.

> This configuration excludes the packages org.lwjgl.egl, org.lwjgl.opengles, org.lwjgl.demo.ovr as well as the class org.lwjgl.glfw.GLFWNativeEGL.java from the build path, as those give compiletime errors when building LWJGL3 with default settings via ant.

This is no longer true since 0919613117450c7e050bfc6a45e8bcc035e68f69.

httpdigest:
Unfortunately, the .project and .classpath files have to be at the project root directory. See: https://bugs.eclipse.org/bugs/show_bug.cgi?id=78438
It is also not possible to use relative paths that use `..` in them. The paths must be absolute or must be a direct child path of the path where the .classpath file resides. Eclipse gives horrible error messages when I altered the .classpath file manually to have relative paths and them imported the project.

Noviv:
I'm not sure if this would be a solution, but couldn't you just add a <code>.nbproject</code> folder? Netbeans would be able to open the project, and Eclipse has the ability to import Netbeans projects. This would also remove the need to copy any files into the root directory.

Netbeans has the option for a free form project so the default Netbeans <code>build.xml</code> can be replaced with the LWJGL build script. Therefore, only the <code>.nbproject</code> is required.

httpdigest:
> ... and Eclipse has the ability to import Netbeans projects.
Are you sure? I've never seen that. Maybe with an additional plugin, though?

Noviv:
Oh man, I'm sorry, I got it the wrong way! Netbeans can import Eclipse, not the other way around. Sorry!

Noviv:
What if you just add <code>apply plugin: "eclipse"</code> to the <code>build.gradle</code> file? According to [this post](http://stackoverflow.com/questions/10722773/import-existing-gradle-git-project-into-eclipse-for-example-hibernate-orm) Gradle will automatically generate Eclipse files.

I haven't worked with Gradle that much, so please tell me if I missed something.

Spasi:
@Noviv The gradle script currently handles artifact deployment only. It does not build the project and depends on ant for it.

@httpdigest Will you update the pull request to remove the EGL/GLES exclusions? Or want me to do it after merging?

httpdigest:
Screwed the git history up with an additional commit... have to figure out how to do it in 1 commit. :)
I do another PR.

PR #91, opened 2015-09-25
Add Eclipse IDE project files

httpdigest:
These include the .project and .classpath files needed for Eclipse Java projects with a minimum working configuration.


httpdigest:
Now without the EGL sources exclusion.

PR #92, opened 2015-10-02
Fix setDebugStreamConsumer for Windows

Bobo1239:
Windows uses CRLF as line separator instead of only CR or LF. This led to only LF getting trimmed and CR remaining which resulted in log messages getting overwritten by the next message.


Bobo1239:
This only works for "normal" LWJGL log messages. OpenGL debug log messages use LF as line separator.

PR #93, opened 2015-10-02
Fix setDebugStreamConsumer for Windows

Bobo1239:
Windows uses CRLF as line separator instead of only CR or LF. This led to only LF getting trimmed and CR remaining which resulted in log messages getting overwritten by the next message.

(Unfortunately, I amended the commit and force pushed instead of rebasing so I have to create a new PR.)


Spasi:
Hey @Bobo1239, thanks for reporting this issue and taking the time to submit a pull request. I have fixed it with a different implementation, in e8dd93fc7897adfb7b4bd8d83c2067bd4bd12cbb.

PR #111, opened 2015-10-30
Fixed eclipse setup notes

Noviv:
This was not changed when you guys added the Eclipse project files, so went ahead and did it!


Spasi:
Thanks!

PR #121, opened 2015-12-04
Fix GL41.glProgramUniform*v auto-size 'count' parameter

httpdigest:
Reported by infinitewrath on the LWJGL forum:
http://forum.lwjgl.org/index.php?topic=6041.0


Spasi:
Thanks!

PR #189, opened 2016-05-02
Fix float/double typo in BufferUtils

kenzierocks:
Doesn't change anything code-wise, but the documentation now matches the code :)


Spasi:
Thanks!

PR #194, opened 2016-06-07
Fixed String.format syntax error

pierr3:
Small syntax error in the String.format output of the API log, causing an exception if there is an OpenGL context error before GL.createCapabilities();


Spasi:
Thanks!

PR #233, opened 2016-09-11
Fix template name

code-disaster:
I was trying to disable Vulkan bindings to reduce time needed for 'compile-templates'. Didn't work out for other reasons (Kotlin compile errors I do not understand), but nevertheless...


Spasi:
Thanks!

PR #239, opened 2016-09-19
Variable argument list callbacks

code-disaster:
This PR adds utility functions to handle `(const char* format, va_list arguments)` style callbacks.

Instead of trying to parse format strings and argument lists with Java, this calls vsnprintf() on the native side, with a user-provided memory buffer. I also added two convenience functions which use stack memory for buffer storage.

So far this has been tested with one of the callback functions in my bgfx binding branch, on Windows 32/64 bit and on OSX.

It's somewhat fiddly to properly handle vsnprintf() and pass around va_arg parameters in a cross-platform compatible fashion, so feel free to moan and/or reject my typecast rampage. ;)

### DIFF
--- a/config/windows/build.xml
+++ b/config/windows/build.xml
@@ -95,6 +95,7 @@
 				<arg value="/I${src.include}\system\dyncall"/>
 				<arg value="/I${src.include}\system\jemalloc" if:true="${binding.jemalloc}"/>
 				<arg value="/I${src.include}\system\jemalloc\msvc_compat" if:true="${binding.jemalloc}"/>
+				<arg value="/wd4710"/>
 				<fileset dir=".">
 					<include name="${src.native}/system/*.c"/>
 					<exclude name="${src.native}/system/lwjgl_malloc.c"/>
--- a/modules/core/src/main/java/org/lwjgl/system/MemoryAccess.java
+++ b/modules/core/src/main/java/org/lwjgl/system/MemoryAccess.java
@@ -97,6 +97,9 @@ final class MemoryAccess {
 	// Returns a new direct ByteBuffer instance
 	static native ByteBuffer newDirectByteBuffer(long address, int capacity);
 
+	// The standard C vsnprintf function
+	static native int vsnprintf(long address, int capacity, long format, long valist);
+
 	/** Implements functionality for {@link MemoryUtil}. */
 	interface MemoryAccessor {
 
--- a/modules/core/src/main/java/org/lwjgl/system/MemoryUtil.java
+++ b/modules/core/src/main/java/org/lwjgl/system/MemoryUtil.java
@@ -1662,7 +1662,7 @@ public final class MemoryUtil {
 	 *
 	 * @param address the string memory address
 	 *
-	 * @return the decode {@link String} or null if the specified {@code address} is null
+	 * @return the decoded {@link String} or null if the specified {@code address} is null
 	 */
 	public static String memUTF8(long address) {
 		return address == NULL ? null : memUTF8(memByteBufferNT1(address));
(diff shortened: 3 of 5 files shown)


config/windows/build.xml
```diff
@@ -90,7 +90,7 @@
 		<mkdir dir="${bin.native}/build"/>
 
 		<!-- CORE -->
-		<build name="lwjgl" dest="${bin.native}/core">
+		<build name="lwjgl" dest="${bin.native}/core" flags="/wd4710">
```
Spasi: This disables all warnings and treat-warnings-as-errors for the entire LWJGL core. It's better to isolate this in `MemoryAccess.c`, using a pragma like:

```c
#ifdef LWJGL_WINDOWS
	__pragma(warning(disable : 4710))
#endif
```

modules/core/src/main/java/org/lwjgl/system/MemoryAccess.java
```diff
@@ -194,6 +197,10 @@ default MemoryTextUtil getTextUtil() {
 			return new MemoryTextUtil();
 		}
 
+		default int vsnPrintf(long address, int capacity, long format, long valist) {
```
Spasi: This can be removed.

modules/core/src/main/java/org/lwjgl/system/MemoryUtil.java
```diff
@@ -1766,4 +1766,105 @@ public static String memUTF16(ByteBuffer buffer, int length, int offset) {
...
+	 *
+	 * <p>The current {@code position} and {@code limit} of the specified {@code buffer} are not affected by this operation.</p>
+	 *
+	 * @param buffer       the output buffer
+	 * @param formatString memory address of the format string
+	 * @param valist       memory address of the {@code va_list} argument list
+	 *
+	 * @return the number of characters written to {@code buffer}, excluding the null terminator, or a negative value in case of error
+	 */
+	public static int memVPrintf(ByteBuffer buffer, long formatString, long valist) {
```
Spasi: I think this is the only method we should keep. It will be used very rarely and it's OK to write some extra code to handle the other overloads.

config/windows/build.xml
code-disaster: Oh, that's been a bad idea then. The problem is that this warning is reported by the linker(!), because of link time code generation I suppose, and so far I didn't find a nice solution which works and is limited to this one function or compilation unit.

code-disaster:
Updated & squashed the commits.

> ... if we could maybe also support functions like bgfx_dbg_text_vprintf.

Yes, I thought about that for a while, but decided to let it go and just use bgfx_dbg_text_printf and String.format() for now. So far I can't think of a use case for such a binding.

Spasi:
@code-disaster, thank you for the PR and the above changes. I experimented a bit and decided on a different approach:

- Special support for `va_list` parameters has been added with 5ed6ca8e44e3cdb7f6a83b9642dcdcb7e733cae8. The `va_list` type is in `GlobalTypes`, so it's available across all bindings.
- Instead of special-case support for `vsnprintf`, bindings to `stdio.h` have been added with e0eb9fac8c663049de58ce459b6e320f1289cc82. You should be able to call `vsnprintf` from `trace_vargs` without issues.
- I have [added tests](https://github.com/LWJGL/lwjgl3/blob/e0eb9fac8c663049de58ce459b6e320f1289cc82/modules/core/src/test/java/org/lwjgl/system/libc/StdioTest.java) that demonstrate how to call vararg functions using dyncall.

Thanks again and I'm eagerly waiting a PR for the bgfx bindings. :)

code-disaster:
Nice! This is both ingenious and a little scary at the same time! ;)

PR #240, opened 2016-09-25
bgfx module

code-disaster:
Following up to http://forum.lwjgl.org/index.php?topic=6315.0, here's a PR to open my doings for your feedback.

I also moved all examples which require assets over to lwjgl3-demos.


Spasi:
The bgfx bindings are now complete, including documentation. There have been changes to several functions and structs, mostly because of missing `AutoSize` modifiers.

kappaOne:
Nice work. There have been substantial changes (module system) and additions (BGFX) to LWJGL since the 3.0.0 release, a lot more than one would expect for a point release, maybe bump the next release version to 3.1 instead of 3.0.1?

Spasi:
> maybe bump the next release version to 3.1 instead of 3.0.1?

Makes sense, yes.

dustContributor:
I noticed only recently there were bgdx bindings, great work dude! :smile:

PR #241, opened 2016-10-01
Update bgfx to API version 23

code-disaster:
I noticed the natives are already built with the latest version. The renderer type enum order has been [changed](https://github.com/bkaradzic/bgfx/commit/f1a8c5f19534302000a48e4e8fdf26776516c1b3), which fails the current demos if they request e.g. the Metal or GL renderers.

I opened an [issue](https://github.com/bkaradzic/bgfx/issues/939) to ask for a means to query the API version, so applications could error early if the bindings API doesn't match. I have a variant working here which uses an opaque variant of bgfx_get_interface(), but this feels a little... dirty.


Spasi:
Thanks!

PR #272, opened 2017-01-29
Use ASM to generate classes at runtime; map jclass to Class; add JNI …

httpdigest:
…ClassToPointer function

Why ASM?
- ASM is 3x faster than Javassist in the current use-case
- the ASM jar is smaller
- the code required to generate a class is less

Why map jclass to Class?
- no need to go through JNI just to lookup a Class by name
- in some cases, we already have the Class instance and need not look it up again by name
- more importantly, this allows to define classes as very light-weight HotSpot Anonymous Classes (those cannot be looked-up by name)

Why ClassToPointer?
- some methods seem to need the jclass as a long (in the ClassbackTest)
- we always already have a reference to the actual Class object, only conversion to long is needed

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -486,7 +486,6 @@
 		<lwjgl.javac srcdir="${src.tests}" destdir="${bin.tests}" taskname="javac: Tests">
 			<classpath>
 				<pathelement path="${bin.core}"/>
-				<pathelement path="${lib}/javassist.jar"/>
 				<pathelement path="${lib}/asm.jar"/>
 				<pathelement path="${lib}/testng.jar"/>
 			</classpath>
@@ -546,7 +545,6 @@
 	<target name="tests" description="Runs the LWJGL test suite" depends="compile-tests, compile-native, -init-runtime">
 		<testng outputDir="${bin.html.tests}" classpathref="runtime.classpath" taskname="Tests">
 			<classpath>
-				<pathelement path="${lib}/javassist.jar"/>
 				<pathelement path="${lib}/asm.jar"/>
 				<pathelement path="${lib}/jcommander.jar"/>
 			</classpath>
@@ -579,7 +577,6 @@
 
 		<java classname="${class}" classpathref="runtime.classpath" fork="true" failonerror="${failonerror}" spawn="${spawn}" taskname="Demo">
 			<classpath>
-				<pathelement path="${lib}/javassist.jar"/>
 				<pathelement path="${lib}/asm.jar"/>
 				<pathelement path="${lib}/testng.jar"/>
 			</classpath>
--- a/build.xml
+++ b/build.xml
@@ -487,6 +487,7 @@
 			<classpath>
 				<pathelement path="${bin.core}"/>
 				<pathelement path="${lib}/javassist.jar"/>
+				<pathelement path="${lib}/asm.jar"/>
 				<pathelement path="${lib}/testng.jar"/>
 			</classpath>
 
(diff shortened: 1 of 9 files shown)


Spasi:
> Why ASM?

Yes, please. `javaassist` was a temporary solution because I haven't done any bytecode generation before and wanted something high level. I actually tried `Byte Buddy` first, which was even heavier and I couldn't find a way to generate native methods...

> Why map jclass to Class?

Nice! It hadn't occurred to me that you could pass a `Class` instance as a `jclass` directly.

> Why ClassToPointer?

This method and the change to `FindClass` are actually illegal JNI. All `jclass` references are local and you cannot return them from JNI code. It works in our simple demos, but you can see it fail in practice if you run the `CallbackTest` demo with `-Xcheck:jni`.

Requesting the following changes:

- Remove `javassist` from dependencies and code.
- Revert change to `FindClass`.
- Wrap `ClassToPointer`'s `jclass` in a `NewGlobalRef`.

httpdigest:
> Revert change to FindClass

Is there a reason why we would want FindClass to actually return a long, when Class/jclass is a supported JNI type and would also allow the GC to track references without having to make a global ref?

Spasi:
> Is there a reason why we would want FindClass to actually return a long, when Class/jclass is a supported JNI type

Because if you wanted a `Class`, you'd use `Class.forName` or whatever. I exposed `FindClass` for the explicit purpose of storing the jclass handle somewhere in native code.

> and would also allow the GC to track references without having to make a global ref?

You cannot use a plain jclass reference outside the JNI function that created it (either via `FindClass` or by passing a `Class` instance). A global reference is absolutely required if the `jclass` handle must be cached somewhere else (because `FindClass` is expensive).

httpdigest:
> You cannot use a plain jclass reference outside the JNI function that created it (either via FindClass or by passing a Class instance). A global reference is absolutely required if the jclass handle must be cached somewhere else (because FindClass is expensive).

That is actually not at all true. A global reference is _only_ required iff the reference to the jclass escapes the JNI method other than _through its return value_. Returning a jclass local reference from a JNI function (or _any_ local reference by the way) is fine.

httpdigest:
> Returning a jclass local reference from a JNI function (or any local reference by the way) is fine.

I should add to that, that this of course only works iff the Java-side return type of the Java method is a Java object type, and not a primitive type such as `long`, which is the reason why you had to put NewGlobalRef to the original FindClass method. Because it returned a primitive `long`.

Spasi:
> That is actually not at all true.

Of course it's legal to have a JNI function that returns `jclass` or any other local reference. But what I said is technically correct: the JNI local reference dies the moment you return it and becomes a Java instance reference. You cannot use the local reference, only the Java instance.

> which is the reason why you had to put NewGlobalRef to the original FindClass method. Because it returned a primitive long.

Which, as I said above, was the entire point. Of course the `ClassToPointer` function is a more efficient way to do the same thing, but for the same reason it again has to use `NewGlobalRef`. Original code that fails with `-Xcheck:jni`:

```c
JNIEXPORT jlong JNICALL ClassToPointer(..., jclass klass) { // klass is a JNI local ref
	return (jlong)(intptr_t)klass; // klass "escapes the JNI method other than through its return value"
} // klass local ref is deleted automatically
```

and fixed:

```c
JNIEXPORT jlong JNICALL ClassToPointer(..., jclass klass) {
	return (jlong)(intptr_t)(*env)->NewGlobalRef(env, klass);
} // klass local ref is deleted automatically, but the global ref is fine
```

Spasi:
Could you please restore the note in the `FindClass` documentation?

Spasi:
I ended up removing both `FindClass` and `ClassToPointer`, because:

- There is no point to call `FindClass`  from Java, when you can simply use a `Class`.
- The updated `ClassToPointer` was effectively identical to calling `NewGlobalRef` with a `Class` parameter.

I've cleaned-up and rebased the tinycc branch to reflect these changes.

httpdigest:
Yes, that's much cleaner. Nice!

PR #279, opened 2017-02-06
Add missing dependency check for org.lwjgl.demo.openal.HRTFDemo

TheMrMilchmann:


TheMrMilchmann:
Hey @Spasi,

I have updated the pull request and removed the (now) unecessary commit. You're indeed right. I was only building a small subset of LWJGLs bindings in order to reduce the build time.

PR #308, opened 2017-05-13
Update the JavaDoc generation

TheMrMilchmann:
With this PR...
- more reliable `@see` and `@since` tag generation is added,
- JavaDoc lines will always begin with a star ("*"). (Note that this does include code blocks!), and
- less blank JavaDoc will be generated.

The changes have been made in order to enhance the readability and quality of generated documentation. However, this does come at the cost that code snippets may not be copied straight from code blocks any longer (what I consider bad practice anyways).

All differences in the generated code can seen [here](https://gist.github.com/TheMrMilchmann/ea6317759b53a1b7ba3703733e97420c).

This also fixes a bug that sometimes caused `@since` tags to not be generated properly ([current](https://github.com/LWJGL/lwjgl3-generated/blob/master/java/org/lwjgl/glfw/GLFW.java#L761) vs. [updated](https://gist.github.com/TheMrMilchmann/ea6317759b53a1b7ba3703733e97420c#file-diff-txt-L1587)).

(A piece of documentation on how commits should be titled would be nice. I hope that this is in compliance with the new format.)

### DIFF
--- a/modules/generator/src/main/kotlin/org/lwjgl/generator/CallbackFunction.kt
+++ b/modules/generator/src/main/kotlin/org/lwjgl/generator/CallbackFunction.kt
@@ -182,13 +182,14 @@ fun String.callback(
     functionDoc: String,
     vararg signature: Parameter,
     returnDoc: String = "",
+    see: Array<String> = emptyArray(),
     since: String = "",
     init: (CallbackFunction.() -> Unit)? = null
 ): CallbackType {
     val callback = CallbackFunction(packageName, className, returns, *signature)
     if (init != null)
         callback.init()
-    callback.functionDoc = { it -> it.toJavaDoc(it.processDocumentation(functionDoc), it.signature.asSequence(), it.returns, returnDoc, since) }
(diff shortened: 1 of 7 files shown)


Spasi:
Good job @TheMrMilchmann, we needed these fixes.

This PR has been merged manually with 2dbfac3ff742ba9dc884db2649da758382f5ca94, c3e5ea3ffb7c7da4f804ced10f1ac341684547e2 and ec09a19c3ca1227ee0567a43ed27065fd090e07f. There are a couple more cases fixed and the code is closer to the original implementation, for performance reasons. It's still not as efficient as I'd like, but it's hard to do the necessary refactoring (`NativeClass::convertDocumentation` is the blocker).

> A piece of documentation on how commits should be titled would be nice.

I'm currently evaluating [Conventional Commits](http://conventionalcommits.org/) (with Angular convention's flexibility) in LWJGL and some other projects. Not going to use the SemVer automation, but I think it's a useful convention. Especially in LWJGL with the dozens of bindings, it makes it easier to quickly identify what each commit is about.

I'll post some guidelines if/when I make the decision to adopt it.

PR #312, opened 2017-05-17
build: make gradlew executable on unix

nedtwigg:
The gradlew script should be executable.  This commit just makes changes the execute bit for gradlew.


nedtwigg:
Gradle, Maven, and Ant all at once!!  The jvm build tool trifecta ;-)

PR #320, opened 2017-06-13
fix(generator): javadoc formatting bugs

TheMrMilchmann:
This fixes an issue that could cause javadoc to have the wrong indent in subclasses. Note that I use "could" here intentionally because LWJGL does not currently create subclasses with `@since` or `@see` tags. However, I decided to open this to "clean-up" the relevant code and to prevent this negatively impacting possible future additions to the library.

(There is no difference in the generated output. Hence I don't see the need to include a diff.)


Spasi:
Thanks!

PR #329, opened 2017-08-07
Fix memory leaks

Atry:


Spasi:
The `APIUtil` functionality is used internally by LWJGL to implement the bindings. The `apiArray` methods specifically are used in this way:

```java
long arrayAddress = apiArray(...);
nativeMethod(..., arrayAddress, ...);
apiArrayFree(arrayAddress, len);
```

i.e. there is no memory leak, the encoded strings are freed by a subsequent call to `apiArrayFree`.

Atry:
Sorry I misunderstood the code.

PR #348, opened 2017-11-13
Fix the signature of JNINativeInterface::ToReflectedField

TheMrMilchmann:
Fixes the signature of JNINativeInterface::ToReflectedField. (The method returned a `java.lang.reflect.Method` instead of a `java.lang.reflect.Field`.)

Also removes an unnecessary semicolon in the LibC mac template.


Spasi:
Thanks!

PR #352, opened 2017-11-24
First HWLOC Additions

Sylvyrfysh:
Brings HWLOC branch up to 3.1.5 and adds beginning types

### DIFF
--- a/modules/lwjgl/openvr/src/generated/java/org/lwjgl/openvr/VREventProgressUpdate.java
+++ b/modules/lwjgl/openvr/src/generated/java/org/lwjgl/openvr/VREventProgressUpdate.java
@@ -1,0 +1,201 @@
+/*
+ * Copyright LWJGL. All rights reserved.
+ * License terms: https://www.lwjgl.org/license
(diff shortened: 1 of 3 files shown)


Sylvyrfysh:
This should be good for the base commit. I'll keep working on adding when I have the time

Spasi:
Hey @Sylvyrfysh,

Do you think you'll be able to continue work on this PR? Other than fixing the current conflicts, the template DSL will also have to be updated to the latest version.

Btw, LWJGL now includes a tool that extracts template definitions from C/C++ headers, including documentation. I would recommend using it to recreate all templates and then adding the necessary modifiers (`AutoSize` etc).

Sylvyrfysh:
Yeah, I'll get back to this and restart what's needed.

Sylvyrfysh:
hwloc 2.0 or old stable 1.11? 2.0 is extremely new, and supports modern processor layouts. 1.11 is the ultrastable branch but may not work as well on newer systems

Spasi:
The latest release (2.0.2) looks stable enough now, I don't see a reason to go with 1.11.

PR #361, opened 2017-12-26
NULL value for more readability

Djasir:
Example:
`long window = glfwCreateWindow(640, 480, "Inception", 0, 0);`

At the moment I'm doing something like that:
`public static final Integer GLFW_NULL = 0;`
`...`
`long window = glfwCreateWindow(640, 480, "Inception", GLFW_NULL, GLFW_NULL);`


TheMrMilchmann:
Such a value does already exist. (See `org.lwjgl.system.MemoryUtil.NULL`)

For convenience I suggest using `import static org.lwjgl.system.MemoryUtil.*;`. This way your method call looks like this: `glfwCreateWindow(640, 480, "Inception", NULL, NULL);`

Djasir:
Thank you very much!

PR #363, opened 2018-01-05
build: make native modules optional at runtime

TheMrMilchmann:
Currently, native modules are required at runtime which is counter-productive when native libraries are meant to be loaded from a separate directory instead of extracting them from a jar.
This PR changes the module declarations of binding to depend "statically" on the native modules (which means that the native modules are required at compile time but optional at runtime). The only regression is that native modules are now no longer visible for modules depending on binding modules. However, this should not be a problem since these modules are empty anyway.


Spasi:
Sounds reasonable, but I'm not sure how it will interact with `jlink`. I'm merging this anyway so that it can be tested more widely (will be available in `3.1.6 build 10`).

Issue #367, opened 2018-01-25
Port Eclipse project to the new directory layout


Spasi:
As of 00e1f52cf95c16b0cea34c83372145532b4745ba the LWJGL codebase has a new directory layout. This has broken the included Eclipse project.

upsonp:
Glad I finally checked the issue tracker... I though it was just me.

I've been playing around with it a bit to see if I can fix the build path issues, but I seem to be missing a libs/testng.jar, which I'm assuming is where the org.testing.* package is coming from and why most of the modules/lwjgl/core/src/test/java/ classes are broken.

The other thing that's showing up broken for me is the `@Nullable` annotation. I'm not sure where that's linked from. Eclipse is telling me to import the com.sun.istack.internal package, but something tells me that's not right and the annotation is defined in one of the other source folders that was refactored.

Lots of other missing and broken imports as well, it'll take me some time to track them all down. I tried to run the build.xml with ant from the root, but it craps out on me when it gets to building the config/linux/build.xml

unfortunately Ant and Gradle are foreign territory for me so I'm not sure how to debug them.

Spasi:
You should first do a successful Ant build before attempting to setup the Eclipse project.

- Download and extract Ant.
- Add Ant's `bin` folder to the path.
- Export/set `JAVA_HOME` to a version 8+ JDK.
- Run `ant compile-templates compile-tests` in lwjgl3's root directory.
- Run `ant tests` if there's a working C/C++ compiler available. _(optional)_

This will download all dependencies and build the project.

upsonp:
`ant compile-templates compile-tests` completed successfully

`ant test` failed in the same place in the linux build as it did before. I think it might be related to the 
`BUILD FAILED
/home/pram/development/eclipse/lwjgl3/build.xml:442: The following error occurred while executing this line:
/home/pram/development/eclipse/lwjgl3/config/linux/build.xml:197: 
The following error occurred while executing this line:
/home/pram/development/eclipse/lwjgl3/config/linux/build.xml:201: exec returned: 1
The following error occurred while executing this line:
/home/pram/development/eclipse/lwjgl3/config/linux/build.xml:83: The following error occurred while executing this line:
/home/pram/development/eclipse/lwjgl3/config/linux/build.xml:33: apply returned: 1`

I did get things working, but probably not in the way you'd like. I added a pom.xml file and used Maven to gather the missing libraries.

[...]

Issue #349, opened 2017-11-19
Port ExcludeDoclet to jdk.javadoc.doclet


Spasi:
Java 10 is going to drop `com.sun.javadoc`.

Java 8 compatibility is not required, LWJGL javadoc is always generated on Java 9 or higher.

Spasi:
Obsolete as of 0d7917de5839436af36fc3908c7b4cb30e410a8c.

PR #373, opened 2018-02-11
That's done it

upsonp:
Related to issue #367 I've updated the eclipse .classpath file.

The steps to set up lwjgl 3 in eclipse are:

1. Check out git project (do not add as existing project)
2. Run ant
- [ ] Open terminal
- [ ] CD to lwjgl3 directory
- [ ] `ant init`
- [ ] `ant compile-templates compile-tests`
3. Copy .classpath and .project from lwjgl3/config/ide/eclipse to the root lwjgl3/ directory
4. In Eclipse (File > Import > Existing Project Into Workspace)
- Select the root lwjgl3/ directory
5. Related to issue #349
- Open Build Path > Libraries tab and add the Tools.jar file as an external jar from your /usr/lib/jvm/\<java implementation\>/libs directory
6. If you have TestNG installed right click the /modules/lwjgl/core/src/test/java folder and Run As > TestNG Test
7. If you're running Linux expect 25 of 30 passes. For some reason the 5 windows specific tests don't work ¯\\\_(ツ)\_/¯


Spasi:
Afaict, the updated classpath includes the core module only. Isn't it possible to add the binding modules as well? (they each have one or more of main, generated and test sources)

upsonp:
It is possible. I may not understand exactly what you're requesting though.

I set up mine with the following as source folders
1. `module/samples/src/test/java/`
2. `module/lwjgl/core/src/generated/java/`
3. `module/lwjgl/core/src/test/java`
4. `module/lwjgl/core/src/main/java`

Then I added all of the `bin/classes/<module>/` folders as part of my class path libraries.

I did so specifically to play with the EasyFont.java class to figure out how fonts work. And it does work for me, I have unrelated questions for you about that at some point. Let's keep this concise.

**I think** the way you want it is set up is so all the
1. `/module/lwjgl/<module>/src/generator/java`
2. `/module/lwjgl/<module>/src/main/java`
3. `/module/lwjgl/<module>/src/test/java`

are all added as ***source*** folders as well. 

The advantage to that is everything is available for development immediately, the down side is it really clutters up the package explorer.

Spasi:
> I may not understand exactly what you're requesting though.

Hopefully, when compiling the project from Eclipse, it should generate classes to the same locations as Ant. I'm not familiar with Eclipse, but this is what it looks like in IDEA:

lwjgl_idea

Each Java module (the LWJGL core module or a binding module) corresponds to an IDEA module. Each IDEA module has its content root and can have multiple source and test folders. The source classes are compiled to one folder (`bin/classes/lwjgl/<module>`), the test classes are compiled to another folder (`bin/classes/test`). Each IDEA module can also have dependencies to other IDEA modules (e.g. the `glfw` module depends on `core`, `egl` and `vulkan`). This helps with isolation (making sure there are no illegal dependencies, etc) and IDE performance.

There are also the Kotlin modules (generator and binding templates) with similar setup. Not sure if the Eclipse Kotlin plugin is any good though.

The samples module depends on all other modules and compiles to `bin/classes/samples`. 

The IDEA project also has run configurations for running the tests with the TestNG plugin and running the code generator. That's just `ant generate` but IDEA builds the Kotlin source before launching the Ant target. This is very useful because it does incremental compilation, which is not available via Ant (a full rebuild takes too long).

upsonp:
I'll digest and think on this for awhile. I'm pretty much out of time for the weekend and have to get a sick daughter to bed. I might have some time tomorrow at some point to work on this some more.

I know I can set the source folders in eclipse to compile to specific output folders, I think that's closer to what you're describing. At the moment everything gets compiled to lwjgl/bin/eclipse rather than to the ant output folders. I'll have to go through and individually set each source folders output.

upsonp:
I've done what I can here. Unfortunately Eclipse doesn't really seem to be set up, or easily set up, to manage projects the way intellj is. What I have seems to work and outputs compiled classes to the proper `bin/generator`, `bin/samples`, `bin/tests` or `bin/lwjgl/<module>`, including the Kotlin classes if you have the Kotlin extension installed in eclipse (Kotlin extension doesn't seem to do package management, just compiles), but this whole set up is really unwieldy for the way I would typically do my development.

The Eclipse TestNG plug-in does pretty well for running the tests (although I prefer jUnit myself)

I'm really not sure how useful this would be to do a full project build from scratch because you can't specify source dependencies and the priority order things need to be built in. Running the Ant build is by far the best way to compile the whole project. I have found I have to run Ant twice, I assume that's because of what @Spasi said about incremental building not available via Ant. In the first run the Kotlin packages are missing, even though they do get compiled out of order with the rest of the build. The second build has the required Kotlin packages already built... I could be way off base on that. I don't typically work with projects this large that have these kinds of interdependencies.

I would probably set a project up to only include the module sources I need to work on at any one time and use link the `bin/lwjgl/<module>` classes as build path Libraries, then use the ant build, onece, to compile the modules into the `bin/lwjgl/<module>` folders for anything I'm not currently working on.

upsonp:
We may want to cut this pull request.

I've been doing a lot or reading up on Java 9 modules after seeing the module-info.java files in the module directories and wasn't really sure what they were about. I'm wondering if there isn't a better way to set this up in Eclipse. When @Spasi was talking about how the modules were set up in IDEA I thought that was a feature specific to IntelliJ, but now I'm wondering if he wasn't using the modules feature in Java 9 (with some IntelliJ support). I do less Java development than I use to and only recently moved to Java 8 so I might need to get caught up on the latest trends. Apparently Eclipse Oxygen 4.7 (which I haven't used) does something with Java 9 and modules and I need some time to research and play with the feature to find out what's possible.

In the meantime, I'll leave it to @Spasi to either close this off or to merge the pull request. It might be a "better than nothing" solution for the moment. Although I think including an Eclipse .classpath file with the bare-bones (just the core and sample modules) and a tutorial on adding sources would be better than what this has become. I've always preferred eclipse (for more than just Java development), but for this type of project setup it's just a mess. I'd rather learn IntelliJ at the moment than develop with the set up I have.

Spasi:
Having a premade IDE project is useful for the following reasons:

- Makes it easy for interested users to explore the codebase.
- Makes it easy to run tests/samples (existing, or modified, or new).
- Makes it easy to develop new bindings.

Only the first two are important in our case. The third one is good to have but not critical. It's OK if we require use of IntelliJ for a good experience, at least until there's a mature Kotlin plugin for other IDEs.

With that said, LWJGL indeed supports Java 9 and its artifacts are explicit JPMS modules. But this is only done in the `ant release` target, in a somewhat hacky way. There is no clean way to develop a Java 8 library and also support Java 9 modules (that I know of, with the current tooling).

However, LWJGL has been modular for a long time, meaning you could [choose](https://www.lwjgl.org/customize) exactly the subset of bindings that are useful to your project. To reflect that (and also the Java 9 module organization), the codebase has been recently changed in such a way that the source of each binding is isolated in its own directory. The output directories are also split, just like the LWJGL jars. The new layout is easier to understand and helps when you want to focus on a specific binding or set of bindings.

Anyway, the IDE setup does not have to support Java 9 modules. The IntelliJ project doesn't either and in fact `module-info.java` files are not even marked as sources. You don't have to organize the bindings in "IDE modules" either, as long as the right paths are added to the classpath and the output directories are configured properly.

Your latest commit has two issues that must be fixed before this PR is merged:

1. The `modules/lwjgl/<module>/src/generated/java` sources must be compiled to `bin/classes/lwjgl/<module>` (just like the `main` sources).
2. The `modules/lwjgl/<module>/src/main/kotlin` sources must all be compiled to `bin/classes/templates`.

upsonp:
Ok, Thanks for clearing that up. I should have realized the `generated/java` should compile to the same class dirs as the `main/java`. Should be an easy fix I'll start on.

Spasi:
No progress, so closing this PR.

PR #374, opened 2018-02-12
Integrate the const modifier into the type system

TheMrMilchmann:
This PR aims at improving readability and fexibility of how native types are represented in templates by bringing them more in line with C conventions. Additionally templates are cleaned up further by removing pointer type definitions from `**Types.kt` files for types that can be accessed by using the `.p` property (e.g. `int_p` for `int.p`).

The `const`-modifiers and `.p`-extensions are replaced by lazy properties (`NativeType::const`, and `NativeType::p` respectively).

**Examples**
`int_p` becomes `int.p`
`const..int_p` becomes `int.const.p`
`const..int_pp` becomes `int.const.p.p`

**Notes**
This should be merged together with LWJGL/lwjgl3-vulkangen#1 which adjusts the the Vulkan template generator accordingly.


Spasi:
The original goal of this PR was fixing the design mistake of making `const` a modifier, instead of having support for it in the type system. Also, `const` is now always rendered to the right side of the type to which it applies, for consistency.

But the real breakthrough came from using Kotlin lazy properties to avoid excessive allocations when applying the new DSL. It led to a major refactoring of the entire type system: it's now simpler, hides implementation details more effectively, provides more type-safety at (kotlinc) compile time.

PR #383, opened 2018-04-11
Separate template sources from main artifact sources

TheMrMilchmann:
Quoted from Slack:

> @ spasi "templates" is an arbitrary name I chose. My opinion/idea is based on the "Introduction to the Standard Directory Layout" in the Maven Guides [1].
>
>Directories located immediately under `src` are referred to as "artifact producing source directories" in the guide and roughly map to `SourceSet`s in Gradle [2]. A simple Kotlin project in Gradle thus expects that all sources under `src/main/java` and `src/main/kotlin` are used for the same artifact. For a typical binding in LWJGL the "`main` build artifact" directory contains three language directories: `c`, `java`, and `kotlin`. While the first two are directly used to build the shipped artifacts (native binaries and binding Jar respectively), the latter contains sources that are only used during the build process and not part of the main artifact. Thus, I think it would be better to move the templates into a separate artifact producing source directory.
>
>[1] https://maven.apache.org/guides/introduction/introduction-to-the-standard-directory-layout.html
>[2] https://docs.gradle.org/current/userguide/java_plugin.html#sec:java_source_sets

Additionally, I noticed that VMA has not been added to the `forEachBinding` macro and fixed this.


Spasi:
Thanks!

PR #398, opened 2018-07-10
fix(GL/GLES): throw an exception for calls without OpenGL context

Pokechu22:
Before, a cross-thread  would cause the JVM to crash (in 3.1.6, without any clear indication as to why).  Now, an exception will be raised instead, which could be handled by the game (e.g. `Thread.setDefaultUncaughtExceptionHandler`) and by default is logged (`Exception in thread "Thread-1" java.lang.IllegalStateException: No LWJGL context found in the current thread.`).

A simple test case, which triggers when `C` is pressed in the space game demo:

```diff
diff --git a/src/org/lwjgl/demo/game/SpaceGame.java b/src/org/lwjgl/demo/game/SpaceGame.java
index ab93c40..982f8cf 100644
--- a/src/org/lwjgl/demo/game/SpaceGame.java
+++ b/src/org/lwjgl/demo/game/SpaceGame.java
@@ -276,6 +276,15 @@ public class SpaceGame {
                 if (key == GLFW_KEY_ESCAPE && action == GLFW_RELEASE) {
                     glfwSetWindowShouldClose(window, true);
                 }
+                if (key == GLFW_KEY_C && action == GLFW_PRESS) {
+                    new Thread() {
+                        @Override
+                        public void run() {
+                            System.out.println("Trying to call glGenTextures cross-thread...");
+                            System.out.println("Result: " + glGenTextures());
+                        }
+                    }.start();
+                }
                 if (action == GLFW_PRESS || action == GLFW_REPEAT) {
                     keyDown[key] = true;
                 } else {
```

The code for this change was already partially added in 622d5a2b41ee9f1e423626ea294a08bbba5c48e6, but it called `functionMissingAbort` in that case which crashes the JVM (although it does also log a warning).  I've cleaned up the change there so that it raises an `IllegalStateException` instead, which should be much more usable; this can happen from a user error and I think it's better to handle it in a user-friendly way.  (`functionMissingAbort`, as far as I can tell, can't happen that way so I think it's fine to leave it as crashing the JVM).

### DIFF
--- a/doc/notes/3.1.7.md
+++ b/doc/notes/3.1.7.md
@@ -68,7 +68,8 @@ This build includes the following changes:
     * Yoga
     * Parameters of struct multi-setter methods
 - OpenCL: Fixed `CL_DEVICE_PREFERRED_VECTOR_WIDTH_INT` constant name.
-- OpenGL/GLES: The thread-local-free function pointer lookups do not depend on JVMTI anymore.  
+- OpenGL/GLES: The thread-local-free function pointer lookups do not depend on JVMTI anymore.
+- OpenGL/GLES: An exception is now thrown when called from a thread lacking an OpenGL context, instead of crashing the JVM.
 
 #### Breaking Changes
 
--- a/modules/lwjgl/core/src/main/c/org_lwjgl_system_ThreadLocalUtil.c
+++ b/modules/lwjgl/core/src/main/c/org_lwjgl_system_ThreadLocalUtil.c
@@ -5,10 +5,17 @@
 #include "common_tools.h"
 
 static jint JNICALL functionMissingAbort(void) {
-    fprintf(stderr, "[LWJGL] No context is current or an unavailable function was called. The JVM will abort execution. Inspect the crash log to find the responsible Java code.\n");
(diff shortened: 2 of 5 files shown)


Spasi:
> I do see that JNI has a FatalError function, which logs the stacktrace and a message and (more importantly) doesn't return but instead kills the JVM.

Actually, that's perfect. The stack trace is better than the one you get in the crash log (includes Java source lines) and is much less intimidating to new users. Changed to `FatalError` with f18cf23f9aa7e5fbf1d2d103cf1cbc48b8bb9ae6.

The `stderr` output can be easily redirected to a file, with or without log4j.

Pokechu22:
Alright, sounds good (I didn't notice that the other one didn't have line numbers and this one does, definitely much better).  I'll close this PR for now.

Yvvw:
Redirecting stderr to a logger via System.setErr does not appear to capture the error thrown from JNI.

Ideally we could have an exception passed through a hookable function in java before calling FatalError in JNI. This would allow both logging and analytic reporting before the JVM ends, without adding the possibility to prevent the inevitable JNI call.

PokeMMO:
Redirecting stderr to a logger via System.setErr does not appear to capture the error thrown from JNI.

Ideally we could have an exception passed through a hookable function in java before calling FatalError in JNI. This would allow both logging and analytic reporting before the JVM ends, without adding the possibility to prevent the inevitable JNI call.

PR #413, opened 2018-09-06
fix(glfw): fix incorrect value for GLFW_HOVERED

kashike:
https://github.com/glfw/glfw/blob/master/include/GLFW/glfw3.h#L817-L821


Spasi:
Thanks!

Issue #414, opened 2018-09-10
Bug in MemoryUtil.memCopy methods


vearp:
### Environment

* LWJGL version: `3.2.0` 
* LWJGL build #: `build 12`
* Java version: `1.8.0_171`
* Platform: `Windows`
* Module: `core`

### Description
There is a bug in MemoryUtil.memCopy methods. The methods checks if the the destination buffer has enough bytes to hold the remaining bytes from source. But the check is checking the opposite, it checks if the source has enough bytes to hold the dest buffer. If the dest buffer is the bigger one, this leads to a thrown exception.

```
   public static void memCopy(ByteBuffer src, ByteBuffer dst) {
         if (CHECKS) {
             check(src, dst.remaining());
         }
         memCopy(memAddress(src), memAddress(dst), src.remaining());
    }
```

should be:
```
     public static void memCopy(ByteBuffer src, ByteBuffer dst) {
         if (CHECKS) {
            check(dst, src.remaining());
         }
         memCopy(memAddress(src), memAddress(dst), src.remaining());
    }
```
This goes for all the variants.

Spasi:
@vearp Thanks!

PR #416, opened 2018-09-23
fix issue #414

blw043:
This pull request applies the submitter's suggested fix to issue #414 (Bug in MemoryUtil.memCopy methods).


Spasi:
Thanks!

PR #424, opened 2018-11-09
docs: move libdive license out of the source directory

TheMrMilchmann:
Minor cleanup to bring this in line with the other modules. Pretty self-explanatory.


Spasi:
Thanks!

PR #427, opened 2018-11-18
Make Listener3i take ALint instead of ALfloat

Dico200:
Bug in lwjgl's bindings for OpenAL 1.1


Dico200:
There might be a problem with this as the corresponding JNI.invoke....(...) method isn't present. Sorry I forgot to mention.

Spasi:
Yes, the generator takes care of it. Try `ant compile-templates generate`.

PR #442, opened 2019-01-26
Relative property location is invalid when building in top-level dirs on linux

layus:
Trying to build lwjgl on linux failed for me because `src.main.rel` was improperly set to `../../modules/opengl/[...]` when it was supposed to be `../../../../modules/opengl/[...]` where the four `..` come from https://github.com/LWJGL/lwjgl3/blob/015c9a95b5ac98fa53e7e5ba359541b12e6798de/config/linux/build.xml#L20

This happens because `src.main.rel` is set with a relative property location in
https://github.com/LWJGL/lwjgl3/blob/015c9a95b5ac98fa53e7e5ba359541b12e6798de/config/linux/build.xml#L79

and finally used after going two directories deeper (`@{dest}`) at https://github.com/LWJGL/lwjgl3/blob/015c9a95b5ac98fa53e7e5ba359541b12e6798de/config/linux/build.xml#L52

Now, if you happen to build in `/build/lwjgl-xxx/`, the relative path will be truncated (because in `/build/lwjgl-xxx/` `../../../..` is the same as `../..` so some kind of canonicalization occurs.)

Proposed solution in this PR. Tested and working by only changing `src.main.rel`. This PR is more a POC than something to be merged as-is.

### DIFF
--- a/config/linux/build.xml
+++ b/config/linux/build.xml
@@ -75,9 +75,9 @@
             <local name="src.main"/>
             <local name="src.main.rel"/>
             <local name="src.generated"/>
-            <property name="src.main" location="${module.lwjgl}/@{module}/src/main/c" relative="true"/>
-            <property name="src.main.rel" location="${module.lwjgl.rel}/@{module}/src/main/c" relative="true"/>
-            <property name="src.generated" location="${module.lwjgl}/@{module}/src/generated/c" relative="true"/>
+            <property name="src.main" value="${module.lwjgl}/@{module}/src/main/c"/>
+            <property name="src.main.rel" value="${module.lwjgl.rel}/@{module}/src/main/c"/>
+            <property name="src.generated" value="${module.lwjgl}/@{module}/src/generated/c"/>
 
             <local name="name"/>
             <condition property="name" value="lwjgl" else="lwjgl_@{module}">


Spasi:
Hey @layus,

I'm not sure what this means:

> if you happen to build in `/build/lwjgl-xxx/`, the relative path will be truncated

The Ant build is meant to be executed from the repository root. What are you doing differently exactly?

layus:
What I am saying is that building from the repository root, when that repository root is not deep enough in the filesystem tree, fails.

For example, assuming the repository root is in a dir named lwjgl-xxx, building with the repository root in /home/user/projects/lwjgl-xxx will work, and building in /build/lwjgl-xxx will not. Most probably, building in /lwjgl-xxx or /var/test/lwjgl-xxx will not. The build is somehow dependent of the (filesystem) depth where the checkout is made.

I of course always start the build in the checkout root.

layus:
Or said differently, that the `../../../..` in `module.lwjgl.rel` was meant to be used in a different cwd than where it is defined. It was not ported correctly in the build system refactoring and was not intended to be used in a `<property name="..." location="..." relative="true"/>`

Spasi:
Thank you @layus!

PR #463, opened 2019-05-02
WIP: build: arm32/arm64 build support.

PokeMMO:
```
#Add support for multiarch environments.
dpkg --add-architecture armhf
dpkg --add-architecture arm64

#install toolchains
apt install gcc-arm-linux-gnueabihf g++-arm-linux-gnueabihf gcc-aarch64-linux-gnu g++-aarch64-linux-gnu

#Afterward you can install dependencies for other archs like
apt install libgl1-mesa-dev:arm64 libgl1-mesa-dev:armhf
```

TODO:
- [ ] cleanup the conditions
- [ ] document required dependencies
- [ ] fix all builds and flags
- [ ] get travis to build arm32/arm64
- [ ] add `CROSSBUILD` or similar flag to enable/disable gXX prefixes.
- [ ] replace suffixes with folder per arch?
- [ ] SharedLibraryLoader support.

### DIFF
--- a/config/linux/build.xml
+++ b/config/linux/build.xml
@@ -5,14 +5,12 @@
 <project name="native-linux" basedir="../.." xmlns:if="ant:if" xmlns:unless="ant:unless">
     <import file="../build-definitions.xml"/>
 
-    <condition property="LIB_POSTFIX" value="" else="32">
-        <equals arg1="${build.arch}" arg2="x64"/>
-    </condition>
+    <property name="gcc.prefix" value="aarch64-linux-gnu-" if:set="build.arch.arm64"/>
+    <property name="gcc.prefix" value="arm-linux-gnueabihf-" if:set="build.arch.arm32"/>
 
-    <condition property="build.arch.x64" value="true" else="false">
-        <equals arg1="${build.arch}" arg2="x64"/>
+    <condition property="gcc.prefix" value="${gcc.prefix}" else="">
+        <isset property="gcc.prefix"/>
     </condition>
-
     <condition property="gcc.suffix" value="-${gcc.version}" else="">
         <isset property="gcc.version"/>
     </condition>
@@ -22,8 +20,8 @@
     <macrodef name="compile">
         <attribute name="dest" default="${dest}"/>
         <attribute name="lang" default="c"/>
-        <attribute name="gcc.exec" default="gcc${gcc.suffix}"/>
-        <attribute name="gpp.exec" default="g++${gcc.suffix}"/>
+        <attribute name="gcc.exec" default="${gcc.prefix}gcc${gcc.suffix}"/>
+        <attribute name="gpp.exec" default="${gcc.prefix}g++${gcc.suffix}"/>
         <attribute name="lto" default="-flto"/>
         <attribute name="flags" default=""/>
         <attribute name="simple" default="false"/>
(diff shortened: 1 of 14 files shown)


config/linux/build.xml
```diff
@@ -5,25 +5,63 @@
...
+       <not>
+          <isset property="LIB_POSTFIX"/>
+       </not>
+    </condition>
+
     <condition property="gcc.suffix" value="-${gcc.version}" else="">
         <isset property="gcc.version"/>
     </condition>
 
+    <condition property="gcc.preffix" value="arm-linux-gnueabihf-">
```
intrigus: Typo. It's 'prefix' not 'preffix'

config/linux/build.xml
```diff
@@ -5,25 +5,63 @@
...
+
+    <condition property="LIB_POSTFIX" value="arm32">
+        <istrue value="${build.arch.arm32}"/>
+    </condition>
+
+    <condition property="LIB_POSTFIX" value="arm64">
+        <istrue value="${build.arch.arm64}"/>
+    </condition>
+
+    <condition property="LIB_POSTFIX" value="">
```
intrigus: In the rest of this file the word suffix is used instead, e.g. Line 63&64.
Might be good to have some consistency.

PokeMMO: Tentative plan (as discussed in slack) was to remove postfix entirely and do per-folder archs. (Similar to android)

config/linux/build.xml
PokeMMO: Thanks

PokeMMO:
Current version will now build all modules except the following: (Edit config/build-bindings.xml to test)

```
    <property name="binding.llvm" value="false"/> <!-- arm32 fails, arm64 fine -->
    <property name="binding.meow" value="false"/> <!-- Missing arm32 values https://github.com/LWJGL/lwjgl3/blob/master/modules/lwjgl/meow/src/main/c/meow_intrinsics.h, arm64 still fails for other reasons -->
    <property name="binding.nfd" value="false"/> <!-- libgtk-3-dev can't be installed for multiple archs... -->
    <property name="binding.openvr" value="false"/> <!-- No arm -->
    <property name="binding.sse" value="false"/> <!-- SSE -->
    <property name="binding.tootle" value="false"/> <!-- JMLSSEVec, hard dep on sse -->

Spasi:
Thank you @PokeMMO!

Summary of final changes:

- Kept a single `build.xml` for all Linux builds. I wanted to avoid duplication and the current version is reasonably clean.
- Cross compilation is enabled with the already supported `LWJGL_BUILD_ARCH` environment variable. E.g. `export LWJGL_BUILD_ARCH=arm64`.
- Shared libraries now live in per-architecture subfolders and are named identically (i.e. dropped the `32` suffix from Windows x86 binaries). This applies to both the Ant build and the natives artifacts for deployment.
- Introduced the `Platform.getArchitecture` API.
- Fixed the LLVM bindings for arm32 (was a generator bug) and meow's build flags for arm64.
- Refactored LWJGL-CI to use a matrix of jobs for Linux & macOS. This makes it much simpler to maintain and add new platforms/architectures in the future. The script that builds `lwjgl3` is [here](https://github.com/LWJGL-CI/lwjgl3/blob/master/.travis.yml).

Issue #428, opened 2018-11-19
Bullet C API first thoughts


dustContributor:
Hi! This is more of an opinion but now that I've been trying to port my stuff to Bullet's C API and worked a bit with it I'm having doubts about it.

The API itself is very awkward to use. I didn't think it'd be a problem but all the indirect handles and indices you have to keep around to operate on things are very annoying.

The API refuses to work with structs directly, since it's thought as a client-server architecture, what you have on the "client side" are usually int64 handles or int32 indices referring to some item on the server side, and a set of many functions that each mutate/read a field of the logical structure on the server side, one by one.

The API is robotics and Python oriented, so I am guessing it only exposes what is useful and practical for both of them.

This means that something really useful in games like collision filters isn't there by default, you have to enable it through a plugin. Moreover, something like a kinematic body (like say, a player) isn't there at all, so instead of setting the velocity of the player you have to move it around with instantaneous forces (unless there is actually a way and I'm an idiot), using forces this way is very hard to fine tune player movement or anything really.

This also means there is a lot of fluff there that isn't relevant for games at all. There are a lot of core functions there purely dedicated to work with the included visualizer of the physics simulation. Until very recently the only way to load an arbitrary mesh body was through a robotics focused file format, instead of you know, an array of vertices.

What I am trying to say is that the client-server layer between the user and Bullet isn't helping at all. It's just bloat a game doesn't needs.

Now the thing is that for 3D there aren't many physics libraries to use out there. And out of them all, Bullet is probably the most documented and widely used, although that's the case for the plain C++ library, not the C API, I doubt there is any game made with it.

It doesn't means there aren't any alternatives:

[Open Dynamics Engine](https://bitbucket.org/odedevs/) has been used on a bunch of games (Techland stuff, ie Call of Juarez, Dead Island) and I've read it has a C API... I just don't understand enough of the repo to know where to look for it. Looking for examples I get results from 2006. It's BSD or LGPL licenced. There is a Java port of it but it uses 64 bit floats everywhere and the first issue you see in the repo is an unresolved high garbage creation complaint from 2015, both of them bad signs.

[...]

PR #480, opened 2019-07-02
 fix(bullet): add missing CreateCollisionShape methods 

Alex-----:
Haven't tested, I did up to `ant compile` in the build process but stopped after that since I don't have a native compiler on this pc.


Alex-----:
I switched to using libbulletc from here https://github.com/AndresTraks/BulletSharpPInvoke which is a more direct c interface to bullet's c++ api. Have only tested it on windows 64 bit so far. Also modified the generate and the extractor a bit.

Spasi:
Hey @Alex-----, great work!

I won't have to review this properly for several days. In the meantime, could you please comment on how it feels to use as an API? Is it much simpler than the current bindings? Also, could you do some performance testing and post a comparison between the two implementations?

Spasi:
Also, do you think it might be better if we combined all the different classes to a single one? I'm thinking that maybe, when the new bindings are used in application code, you'd need to have too many `import static` statements to do anything useful.

Spasi:
I've updated the PR with significant changes and fixes. There's no generated native code now (the bindings use `org.lwjgl.system.JNI`) and the libbulletc shared libraries are downloaded from LWJGL's CI (which now has [libbulletc builds](https://github.com/LWJGL-CI/BulletSharpPInvoke) for all platforms). I've also updated `HelloBullet`, it's now a port of [this test](https://github.com/LWJGL-CI/BulletSharpPInvoke/blob/master/libbulletc/test/bulletctest.c). Successfully tested on Win x86/x64 and Linux.

There's still work to be done, mainly the correct definition of all callback function types. See `btActionInterface.kt` for an example of how it should look.

Alex-----:
@Spasi https://github.com/LWJGL-CI/BulletSharpPInvoke/blob/master/libbulletc/src/main.h has some defines at the top that disable parts of it, these should probably be removed. I've only used the first one so far (BULLETC_DISABLE_IACTION_CLASSES) for the btKinematicCharacterController.

I haven't done a lot with it yet but it does seem much simpler than the previous bindings, the documentation for the c++ api is much better. When I was trying to use the previous bindings before I had to keep reading the source to see which arguments I needed to call the bullet c++ api methods that I wanted to call.

We could combine them all into one class, another option would be to remove the class name prefixes from the methods and not use static imports.

Alex-----:
I added libvpx bindings. I'm using `../configure --disable-shared --target=x86_64-win64-vs16 --disable-examples --disable-tools --disable-docs --disable-unit-tests --enable-static --enable-static-msvcrt --enable-pic` to build it. I haven't updated to the latest version of LWJGL yet though.

Spasi:
Closing this, see [this comment](https://github.com/LWJGL/lwjgl3/issues/428#issuecomment-948024976) for details.

Thanks for the effort @Alex-----!

PR #484, opened 2019-07-13
Update par_shapes and added par_streamlines bindings

TheMrMilchmann:
Really not much else going on here.

I essentially voided `__attribute__(unused)` macros on anything other than GCC to avoid compilation errors on windows. Not sure if that is the best fix, but it is certainly the best I could come up with at the moment.

I have no idea if the code currently compiles on macos as I have no device I could test this with.


Spasi:
Thank you @TheMrMilchmann!

PR #492, opened 2019-08-08
Convert publish Gradle script to Kotlin DSL and the maven-publish plugin

TheMrMilchmann:
The old `maven` plugin is about to be deprecated in Gradle 6.0. The [`maven-publish`](https://docs.gradle.org/current/userguide/publishing_maven.html) plugin should be used instead.

Additionally the script has been converted to use Gradle Kotlin DSL. However, until #16 is completed, the benefits are [limited](https://docs.gradle.org/current/userguide/kotlin_dsl.html#kotdsl:limitations).


Spasi:
Thank you @TheMrMilchmann!

I've made some minor adjustments, mainly to make the diff with the Groovy version simpler. The extra commit helps `git log` with following history of the renamed files.

Issue #481, opened 2019-07-05
Please ship a bill of materials dependency


sixcorners:
A bill of materials or bom dependency is one that can be used to keep the versions of dependencies in sync. Instead of adding a bunch of dependencies and setting them all to version 3.2.2 or using properties you could instead add a bom with version 3.2.2 and omit the versions on the dependencies managed by the bom. It's a good way to not having to repeat yourself if there are a bunch of versions that all need to be set to the same thing. It also makes sure you are using a set of versions that have all been tested to work together. It would be nice if lwjgl provided a lwjgl-bom dependency I could import.

Here are a few examples of them:
https://github.com/apache/logging-log4j2/blob/log4j-2.12.0/log4j-bom/pom.xml
https://github.com/FasterXML/jackson-bom/blob/jackson-bom-2.9.9/pom.xml
https://github.com/quarkusio/quarkus/blob/0.18.0/bom/pom.xml
https://github.com/spring-projects/spring-data-build/blob/2.1.9.RELEASE/bom/pom.xml
https://github.com/spring-projects/spring-boot/blob/v2.1.6.RELEASE/spring-boot-project/spring-boot-dependencies/pom.xml

Here is how I would use one in maven: https://logging.apache.org/log4j/2.x/maven-artifacts.html#Bill_of_Material
Here is how to do it in gradle: https://docs.gradle.org/current/userguide/managing_transitive_dependencies.html#sec:bom_import

XenoAmess:
The reason not to do so I guess is people usually not use all of those packages.

TheMrMilchmann:
You don not need to use all of the packages with a BOM. Instead, when you include the BOM, you still need to declare all of the dependencies but not their versions. So, what this does is effectively allowing you to skip declaring the LWJGL version `{numberOfLWJGLDependencies}` times by declaring one additional dependency.

Instead of:
```kotlin
dependencies {
    // define dependencies with versions
    implementation("org.lwjgl", "lwjgl", lwjglVersion)
    implementation("org.lwjgl", "lwjgl-glfw", lwjglVersion)
    implementation("org.lwjgl", "lwjgl-opengl", lwjglVersion)
    implementation("org.lwjgl", "lwjgl-stb", lwjglVersion)
    runtimeOnly("org.lwjgl", "lwjgl", lwjglVersion, classifier = lwjglNatives)
    runtimeOnly("org.lwjgl", "lwjgl-glfw", lwjglVersion, classifier = lwjglNatives)
[...]

PR #494, opened 2019-08-17
Provide a Bill of Materials (BOM)

TheMrMilchmann:
Adds a new publication that defines a Bill of Materials (BOM). This BOM can be consumed as demonstrated [here](https://github.com/LWJGL/lwjgl3/issues/481#issuecomment-517913593).

The coordiantes for the BOM currently are ~`org.lwjgl:lwjgl-dependencies` which introduces a **minimal** risk of name-clashes should LWJGL ever add bindings to a library called `dependencies`. (`org.lwjgl:lwjgl-bom` would also be a good choice in my opinion.)~ `org.lwjgl:lwjgl-bom` since this seems to be the most common choice (as pointed out below).


httpdigest:
I vote in favor of using `org.lwjgl:lwjgl-bom`, since the "-bom" artifact name suffix seems to be pretty much the de-facto standard for such BOMs. See other projects like Spring Framework (org.springframework:spring-framework-bom), or Jackson (com.fasterxml.jackson:jackson-bom), or Jetty (org.eclipse.jetty:jetty-bom) or Netty (io.netty:netty-bom), or Reactor (io.projectreactor:reactor-bom).

TheMrMilchmann:
I fully agree. Updated the artifact name and also added a note in the changelog.

Spasi:
Thank you @TheMrMilchmann!

PR #501, opened 2019-09-02
config for building on freebsd

michal800106:


Spasi:
Thank you @michal800106, LWJGL now has FreeBSD support.

PR #508, opened 2019-09-25
feat(spvc) add SPIRV-Cross bindings

httpdigest:
Due to the lack of function/parameter documentations in the [original C header file](https://github.com/KhronosGroup/SPIRV-Cross/blob/master/spirv_cross_c.h) the generated Java methods/parameters are lacking those as well.


skalarproduktraum:
This is awesome! Any idea when it's ready to be merged?

Spasi:
Thank you @httpdigest!

PR #510, opened 2019-10-12
small refines

XenoAmess:


Spasi:
Hey @XenoAmess,

Could you please squash the 7 commits and update the PR? Also, could you explain what problem these changes are solving?

XenoAmess:
> Hey @XenoAmess,
> 
> Could you please squash the 7 commits and update the PR? Also, could you explain what problem these changes are solving?

squash done.
explaining about this commit is added in the new commit.

modules/generator/src/main/java/org/lwjgl/generator/util/ModuleInfoGen.java
```diff
@@ -67,15 +67,15 @@ public static void main(String[] args) throws IOException {
         }
 
         @Override public boolean equals(Object other) {
-            return name.equals(((Module)other).name);
+            return other != null && (other instanceof Module) && Objects.equals(name, ((Module)other).name);
```
kenzierocks: `other != null` is redundant with `other instanceof Module`, which already returns `false` when `other` is `null`.

httpdigest:
I think Spasi's question was more in the sense of "did you actually experience an issue/bug when running the code and your changes actually fix such observed bug" and not so much of a theoretical "it looks like we should do a null-pointer check here" motivation.

XenoAmess:
> I think Spasi's question was more in the sense of "did you actually experience an issue/bug when running the code and your changes actually fix such observed bug" and not so much of a theoretical "it looks like we should do a null-pointer check here" motivation.

Haven't meet any bugs about it.
I just see it when learing about the codes.

modules/generator/src/main/java/org/lwjgl/generator/util/ModuleInfoGen.java
XenoAmess: exactly.

PR #520, opened 2019-12-26
Specifying GLFW 3.2 for MacOS X compatibility

amatiushkin:
Demo sample for gears does not run in MacOS X and gives will known [error](https://stackoverflow.com/questions/20264814/glsl-version-130-on-mac-os-x-causes-error).

It can be fixed by explicitly specifying version as suggested in  [forum](http://forum.lwjgl.org/index.php?topic=5578.0) thread.
```
glfwWindowHint(GLFW_CONTEXT_VERSION_MAJOR, 3);
glfwWindowHint(GLFW_CONTEXT_VERSION_MINOR, 2);
```


Spasi:
Hey @amatiushkin,

Thank you for the pull request! The demo indeed was running fine on GL3+ and your fix solved the issue on macOS. However, I preferred to fix the GL2 implementation instead, so that it can run on older systems as well.

amatiushkin:
@Spasi 

Thanks for the fix, it works.

PR #532, opened 2020-02-13
docs(opengl): add a missing pname

Mouse0w0:
add a missing pname GL_TEXTURE_BORDER_COLOR to glSamplerParameteri


octylFractal:
This should be added to the generator, not to the generated code, e.g. at https://github.com/LWJGL/lwjgl3/blob/master/modules/lwjgl/opengl/src/templates/kotlin/opengl/templates/GL33Core.kt#L117.

Mouse0w0:
OK. I've added it to the generator.

Spasi:
This parameter list is being used in the `(Get)SamplerParameter*` functions. The `TEXTURE_BORDER_COLOR` parameter is already added separately in the `iv` variants, but not in the `i` one. That's because `TEXTURE_BORDER_COLOR` requires 4 values to be specified, so it's not applicable to the single value variant.

(note: `TEXTURE_BORDER_COLOR` as well as `TEXTURE_LOD_BIAS` should be added to the `f(v)` variants)

In general, indeed, the OpenGL bindings javadoc has several issues and a ton of missing information. It'd be a huge undertaking to make it complete & fully correct, at least without an automated solution like the one we have for Vulkan. I will accept PRs that make improvements, however I will personally not contribute anything other than the basics and I don't wish for anyone else to go through the pain of doing it right.

PR #536, opened 2020-02-29
for Callback.Container classes, when it free(), I think it is better  to free its delegate first.

XenoAmess:
for Callback.Container classes, when it free(), I think it is better to free its delegate first.


Spasi:
Hey @XenoAmess,

This would result in a double free. The function pointer stored in `Container` is identical to `delegate.address()`.

XenoAmess:
Oh I see... thx.

PR #548, opened 2020-04-12
fix(core): handle trailing newlines in api version parsing.

PokeMMO:
I've received a crash log from a user unable launch due to a failure in parsing the api version due to a trailing newline character.

The user is running on macOS, on older hardware.

```
java.lang.IllegalArgumentException: Malformed API version string [2.1 NVIDIA-10.0.35 310.90.10.05b12
]
    at org.lwjgl.system.APIUtil.apiParseVersion(Unknown Source)
    at org.lwjgl.system.APIUtil.apiParseVersion(Unknown Source)
    at org.lwjgl.opengl.GL.createCapabilities(Unknown Source)
    at org.lwjgl.opengl.GL.createCapabilities(Unknown Source)
```


Spasi:
Thanks for the PR! I ended up fixing the issue in the regex, instead of trimming the input string.

PR #565, opened 2020-06-07
build: update JOML to 1.9.25

httpdigest:


apostolos:
@Spasi let me know when the files have been uploaded so we can merge this

httpdigest:
It seems that there is nothing to upload. The update-mvn ant target downloads directly from Maven Central, which JOML 1.9.25 has already been published on: https://mvnrepository.com/artifact/org.joml/joml/1.9.25

PR #580, opened 2020-08-17
Don't publish dependency scopes in BOM

TheMrMilchmann:
As discussed, this PR takes care of removing dependency scopes from the BOM. It turns out that this was already fixed in Gradle 6.2, thus I only had to remove one line in the workaround for native artifacts.


Spasi:
Thanks!

PR #627, opened 2021-03-07
New bindings to OpenXR

Sorenon:
Just as a foreword this is my first time creating a pull request this large so apologies in advance if there are any glaring issues that I missed.

### OpenXR is a new cross platform and generic XR api that can be treated as the [replacement for OpenVR](https://steamcommunity.com/games/250820/announcements/detail/2522527900755718764)

Using a tweaked version of vulkangen (which I will create a pull request for after cleaning it up) I have generated functional bindings to the OpenXR api. Currently the templates lack documentation since I wanted to get a working product ASAP. I have also created a functional demo which requires an OpenXR runtime and a stereo headset.

Linux / MacOS support is very limited as I don't have the capacity to run either of those at the moment, but hopefully it should be as simple as adding the needed natives and a few lines of code to the sample.

If you want to get a better grasp of the api [this](https://github.com/ReliaSolve/OpenXR-OpenGL-Example) is a well documented demo of a working OpenXR app in c++

### Natives:
Although the OpenXR runtime is a shared library similar to vulkan-1.dll the intended way to use the library is using the OpenXR-Loader. In the case of LWJGL it should be bundled into a jar file similar to glfw.dll. The OpenXR-Loader takes care of locating and loading the selected runtime, then provides all of the needed functions such as xrGetInstanceProcAddr.
Currently the binaries have to be provided manually at `bin/libs/native/windows/x64/org/lwjgl/openxr/openxr_loader.dll` but I expect that they can be treated the same as the glfw binaries. That is having them compiled remotely on a server and then downloaded when needed.

Precompiled OpenXR-Loader binaries for windows can be downloaded from [here](https://www.nuget.org/packages/OpenXR.Loader/)
And the OpenXR-Loader source and build scripts can be found [here](https://github.com/KhronosGroup/OpenXR-SDK)

### Runtime:
If you want to use the bindings then you will need to have the OpenXR runtime installed. Oculus devices should come with one preinstalled otherwise [this](https://store.steampowered.com/news/app/250820/view/2396425843528787269) describes how to install them with Steam VR


modules/lwjgl/openxr/src/generated/java/org/lwjgl/openxr/XR10.java
```diff
@@ -0,0 +1,1577 @@
...
+
+    // --- [ VK_VERSION_MINOR ] ---
+
+    /**
+     * Extracts the API minor version number from a packed version number.
+     *
+     * @param version the OpenXR API version
+     */
+    @NativeType("uint64_t")
+    public static long VK_VERSION_MINOR(@NativeType("uint64_t") long version) {
```
knokko, line 1373: I think this should be XR_VERSION_MINOR instead of VK_VERSION_MINOR (and the same holds for VK_VERSION_PATCH)

Sorenon: Noted, I'll rename these once I have more time on my hands

modules/samples/src/test/java/org/lwjgl/demo/openxr/HelloOpenXR.java
```diff
@@ -0,0 +1,835 @@
...
+//                        graphicsBinding.set(
+//                            KHROpenglEnable.XR_TYPE_GRAPHICS_BINDING_OPENGL_WAYLAND_KHR,
+//                            NULL,
+//                            GLFWNativeWayland.glfwGetWaylandDisplay()
+//                        );
+//                        this.graphicsBinding = graphicsBinding;
+//                    } else {
+//                        throw new IllegalStateException();
+//                    }
+                    throw new IllegalStateException();
```
knokko, line 267: It would be nice if we could fix the linux support for the OpenGL example before finishing the pull request. Unfortunately, I have never really worked with windowing libraries directly, so I wouldn't know how to do this either. If I have time left, I will give it a try. Also, I think it would be better to create some static utility method that handles this platform-specific case distinction rather than letting users copy this code into all of their projects.

knokko:
If I understand the OpenXR loader correctly, a precompiled loader is only necessary for Windows. In other platforms, the OpenXR runtime should install this system-wide. The bindings seem to be working on Ubuntu already (using the Monado OpenXR runtime). (But not your OpenGL example because of the review I did.) As far as I know, there is no OpenXR runtime for MacOS. Also, OpenXR has graphics support for OpenGL, Vulkan, and Direct3D, but not Metal, so I wonder how such a runtime would work anyway.

I am currently working on a Vulkan example at https://github.com/knokko/lwjgl3-openxr/blob/knokko-dev/modules/samples/src/test/java/org/lwjgl/demo/openxr/HelloOpenXRVK.java . I think it would be good to finish this before merging the pull request.

modules/samples/src/test/java/org/lwjgl/demo/openxr/HelloOpenXR.java
Sorenon: I've done a bit of research into getting XLib support working for the OpenGL and I've come to the conclusion that there are two ways to get it to work. 
* Ditch GLFW entirely in the example (for XLib) and create the window and OpenGL context manually
* Update GLFW to expose some of the needed parameters

While the first solution will most likely work it feels much too hacky for LWJGL. However, the other will require updating both GLFW and the GLFW bindings so will take much longer to be approved despite being a much better long term solution.

knokko: The second option sounds best to me. Why would that take much more time? Not many extra functions are needed, right? (At least, nothing compared to the number of functions this pull request will add for OpenXR.)

I dislike the first option because I would like to minimize platform-dependent code. Such code is nasty because testing it requires checking for each platform.

Spasi:
Hey @Sorenon, I've recently refactored `lwjgl3-vulkangen` to use the Asciidoctor Extensions API (instead of the hacky parsing it was doing with regular expressions). This made parsing a lot more stable and fixed tons of issues in the Vulkan bindings documentation. You may want to rebase your tweaked version for OpenXR and re-generate the bindings + update this PR.

Brettb-gpu:
Having OpenXR support would be a major leap forward for LWJGL. Are there any updates on the status of this PR?

Sorenon:
I've made some headway on my local branch which isn't ready to be pushed yet

Spasi:
Hey @Sorenon,

I've had some free time over the holidays and went ahead with the port of `lwjgl3-vulkangen` to [lwjgl3-openxrgen](https://github.com/LWJGL/lwjgl3-openxrgen). Obviously this replicates most of the hard work you've put into this PR, but I just want to say a massive thank you! Being familiar with how it works, I know it must have been painful for you and be sure that it was an inspiration that someone would do this in order to get a new module into LWJGL. It's motivating knowing that users are really interested and will actually use these bindings.

The OpenXR bindings added with 0fd539de3fb8f43258fe0c97431ac681b5361550 are very similar to this PR, with the following differences:

- Full documentation, similar to the Vulkan bindings. `openxrgen` is currently synced with all the latest features/fixes in `vulkangen`.
- The dispatchable handle classes, other than `XrInstance`, are auto-generated.
- There is no JNI code generated for the two functions in `EXT_conformance_automation` with struct parameters passed by-value. This utilizes the recently added support for libffi downcalls in the Generator and avoids having to bundle an extra shared library, just for those two functions.
- There is also no OpenXR loader library bundled. I think we should wait for the ecosystem to expand a bit first, before deciding if it's necessary or not. For example, we've never had to bundle a loader for OpenCL or Vulkan so far. The `XR` class currently looks for `openxr-loader.dll` on Windows and `libopenxr_loader.so` on Linux, as per the official guidelines.

Obviously, I'd be very happy to receive more contributions from @Sorenon, @knokko or anyone else interested. Looking for:

- Testing! Does OpenXR work with the current bindings?
- Sample code! E.g. it would be nice if you could rebase this PR to the current head and submit the `HelloOpenXR` demo. Btw, the `XRHelper` class should be moved to the samples module.
- Feedback! E.g. let me know if you have any thoughts about the OpenXR loader situation.

Thanks again!

Sorenon:
Well first of all I'd like to apologize for not finishing the PR myself, nevertheless I am very glad to see OpenXR support officially in LWJGL. I have a fairly strong understanding of OpenXR so I can give some responses to your requests. 

Testing: There is already a diverse userbase for the bindings with https://github.com/Sorenon/MCXR and I will put it to work as a testing ground for these improved ones.  

Samples: Once I've validated the bindings works with MCXR I'll try porting the old demos, though I was having a few issues with HelloOpenXR GL (which was part of the reason the PR stalled).  

Loader Feedback: For Linux leaving the loader unbundled makes sense as it can be easily installed through the package manager. However for Windows I think it should be bundled as unlike Vulkan and OpenCL there won’t be a system-wide OpenXR loader dll for the foreseeable future. So the lack of a bundled loader will require developers to deal with the dll file manually.

Spasi:
> Loader Feedback: For Linux leaving the loader unbundled makes sense as it can be easily installed through the package manager. However for Windows I think it should be bundled as unlike Vulkan and OpenCL there won’t be a system-wide OpenXR loader dll for the foreseeable future. So the lack of a bundled loader will require developers to deal with the dll file manually.

Developers having a little bit of trouble is OK. They can easily download/build the generic loader and there are many ways to make LWJGL find it. I'm mostly thinking about end-users. Will OpenXR applications always want to bundle a loader? Or would end-users, via the drivers/runtime of their XR hardware, have one available already?

SWinxy:
Oh man this is awesome! I will try it out when the build becomes available. I don't know how these loaders work, but I assume that if you had an HMD, you're not not going to have Steam, which would install the loader for you?

knokko:
> > Loader Feedback: For Linux leaving the loader unbundled makes sense as it can be easily installed through the package manager. However for Windows I think it should be bundled as unlike Vulkan and OpenCL there won’t be a system-wide OpenXR loader dll for the foreseeable future. So the lack of a bundled loader will require developers to deal with the dll file manually.
> 
> Developers having a little bit of trouble is OK. They can easily download/build the generic loader and there are many ways to make LWJGL find it. I'm mostly thinking about end-users. Will OpenXR applications always want to bundle a loader? Or would end-users, via the drivers/runtime of their XR hardware, have one available already?

I wouldn't want to let developers worry about this. One of the things I always liked about LWJGL is that I only had to worry about *writing code* rather than dealing with the native libraries. I would like to keep the loader until we are sure that we don't need it. Once we are sure that we don't need it, we can simply remove it from LWJGL without anyone else needing to know/worry about this.

**Regarding the `XRHelper`:** 
I would rather not move it to the sample modules because I expect that nearly every XR application will need it. Moving it to samples will prevent applications from depending directly on it, which requires them to copy the code. This sounds like unneeded code duplication to me. 

Also, the `GraphicsBindingOpenGL` stuff in `XRHelper` needs to handle each operating system differently (and even distinguish between X11 and Wayland on Linux). Such code is generally hard to test since it must be tested on each platform. I think it would be better to let LWJGL handle this nasty stuff rather than including such code in actual applications. Furthermore, this implementation is not even finished yet: it can't be completed until https://github.com/glfw/glfw/pull/1925 is merged. If we keep this in LWJGL, we can simply update LWJGL when/if that merge request is merged (and applications only need to upgrade their LWJGL version). If we would keep it in the samples and let applications copy it, *every single* application would have to update it manually when the time comes. When they don't, their applications will work on lesser Linux devices, which is annoying for Linux end users.

**Testing**
Other than that, I will test the new bindings when I get the chance. Thanks for finally finishing this merge request!

knokko:
@Spasi I added the examples that @Sorenon and me made for this merge request and I fixed some bugs that I encountered while doing so. I pushed my changes to https://github.com/knokko/lwjgl3/commits/origin/openxr-samples . I did not create a formal pull request because you probably don't want to copy *all* my changes to the actual LWJGL repository.

## Critical epic bugfixes
That having said, I would really like you to cherry pick the "Critical epic bugfixes" commit.

## The `next` field should be `null`able
I also made some commits that remove the `null` checks for the `next` fields of some OpenXR structs. I do *not* want you to cherry pick these: I would like you to adapt the generator such that the `next` fields of OpenXR stricts are no longer required to be non-`null` (they should be `null`able). I removed the `null` checks manually for the sake of testing and running the examples.

## Dependencies
I also made the commit "Fix dependencies for the OpenXR bindings". I don't know whether you should cherry pick this: I needed it to run the examples in IntelliJ, but maybe I just made some mistakes with the project setup. Use your own discretion here.

## Examples
I also made some commits that add and refactor the OpenXR + Vulkan and the OpenXR + OpenGL example. Both examples compile on my machine and the Vulkan example runs correctly on my machine. I can run the OpenGL example on my machine as well, but that example doesn't support Linux yet (OpenXR + OpenGL + Linux is currently a bit of a nightmare and I'm afraid I can't fix this until the GLFW pull request I mentioned is accepted). @Sorenon You mentioned there was some issue with the OpenGL example. Is that related to this? Or is it something else?

## XRHelper
Like I said in my previous message, I would prefer adding XRHelper to the OpenXR module rather than adding it to the OpenXR sample, which I did on my branch as well. If you really disagree, you can move it yourself.

Spasi:
Hey @knokko,

> Critical epic bugfixes

Thanks!

> The next field should be nullable

This is a bug in OpenXR-Docs, `xr.xml` is missing the `optional` attribute on all `next` members. Vulkan-Docs doesn't have this problem. Added a workaround in `lwjgl3-openxrgen` for now.

> Dependencies

Thanks, incorporated with a few changes.

> Examples

I've refactored a lot of code in the demos, but it was mostly cosmetic changes to follow the `lwjgl3` code style, unless I screwed something up by mistake. I did make one functional change though, the second `vkAllocateMemory` call was using a wrong `VkMemoryAllocateInfo` (line `988` of the original code, `aiSharedMemory` instead of `aiDeviceMemory`). Anyway, please let me know how it works.

> XRHelper

Yeah, moved it to the samples module. The code in there isn't so critical or difficult (well, compared to what you have to do to even get started with OpenXR + GL/VK), but even if it were, there should not be any coupling between OpenXR and GLFW.

Finally, I've bundled the OpenXR loader. Building it for all Linux + Windows architectures we support at [LWJGL-CI/OpenXR-SDK](https://github.com/LWJGL-CI/OpenXR-SDK/actions).

knokko:
Thanks for pointing out the problem with `aiSharedMemory` and `aiDeviceMemory`. Nice catch!

## Testing results
The Vulkan example works on my real HMD when it is connected to my Windows PC. It also works on the virtual HMD on my Linux laptop (I can't connect my real HMD to my Linux laptop, but this has nothing to do with LWJGL).

The OpenGL example doesn't work on the virtual HMD on my Linux laptop (because the OpenGL example doesn't support Linux, and adding support won't be trivial). Unfortunately, it doesn't work on my real HMD either. It looks like this is the problem that Sorenon mentioned. I will try to fix this in the near future.

## Build configurator
I noticed that OpenXR is not yet available in the build configurator, not even in the Snapshot build. Could you add it there? (The current OpenGL example is broken, but this is a snapshot after all, so users know that it might be broken.)

Spasi:
Hey @knokko,

> Build configurator

It was just updated, OpenXR is there now.

knokko:
I fixed the OpenXR + OpenGL example and pushed the final version to https://github.com/knokko/lwjgl3/blob/origin/fix-xr-gl-sample/modules/samples/src/test/java/org/lwjgl/demo/openxr/HelloOpenXRGL.java . To see the actual fixes, check https://github.com/knokko/lwjgl3/commit/210d1df592d1790e32d3cb0ca83b6e77a2d72187 .

When fixing the example, I noticed 2 things that we might want to change in the binding:
- It is currently not possible to change the `type` field of `XrSystemProperties`. I had to work around this by using `memPutInt(systemProperties.address(), XR_TYPE_SYSTEM_PROPERTIES);`. I think it would be better if the `type` field were mutable.
- Setting `XrFrameEndInfo.layers` to `null` does **not** set `.layerCount` to 0, but setting it to non-`null` **does** set the `layerCount` to the right value. This can create unexpected behavior if the `XrFrameEndInfo` was allocated using `malloc` since this causes the value of `layerCount` to be unspecified in case `layers` is `null` (which is sometimes the case). I worked around this by using `.layerCount(didRender ? layers.remaining() : 0)`, but I think it would be better if this wouldn't be necessary.

Spasi:
Hey @knokko,

> I fixed the OpenXR + OpenGL example and pushed the final version

Thanks, it's been merged!

> It is currently not possible to change the type field of XrSystemProperties.

Thanks, fixed this issue with the `type` & `next` members in read-only structs. Also added typed overloads for `next` members, that were also missing.

> Setting XrFrameEndInfo.layers to null does not set .layerCount to 0, but setting it to non-null does set the layerCount to the right value.

I know what you mean. This was a big headache with Vulkan too and in general when designing the API for structs. The problem is that there is a legitimate case where the "size" member could be set explicitly, but the "auto-sized" member is `NULL` anyway. In Vulkan-Docs, it looks like this:

```
<member optional="true"><type>uint32_t</type><name>fooCount</name></member>
<member optional="true" len="fooCount">const <type>VkFoo</type>*<name>pFoo</name></member>
```

When `pFoo` does not have the `optional="true"` attribute, it's a normal auto-sized member and setting `pFoo` to `NULL` would also set `fooCount` to `0`. However, when the attribute exists, we have the case described above. If passing `NULL` did set `fooCount` to `0`, it would break code like this:

```
myStruct
    .fooCount(1)
    .pFoo(NULL);
```

I also dislike that this corner-case exists, but we couldn't figure out a better solution.

Anyway, it seems that `xr.xml` in OpenXR-Docs needs more work and such issues will probably be fixed as OpenXR matures. E.g. in this case, I don't see any reason to have `optional` on the `layers` member. Vulkan-Docs also had a ton of issues, maybe until version 90 or so, that I had to painstakingly workaround to generate the bindings. I hope it won't take that long with OpenXR! :)

knokko:
Thanks for fixing the first issue and explaining the second issue. I don't know a solution for the second issue either, so I guess the example will have to be like this for now.

I think the only thing left to do is fixing the OpenXR + OpenGL Linux support (I know that Linux is not common for VR, but it really annoys me anyway). I will ask for advice on the Monado team for this and improve the example if I can.

knokko:
My OpenGL Linux attempts weren't fruitful, so I will give up on that for now. However, I did find some other things that you might want to improve:
- Make `XrApiLayerProperties.type` mutable. Currently, the example works around this problem using `XRHelper.prepareApiLayerProperties`, but I find it a bit silly that this work-around is needed.
- Do the same for `XrExtensionProperties`.
- Currently, `xrEnumerateSwapchainImages` only accepts `XrSwapchainImageBaseHeader.Buffer` as last parameter. This is inconvenient since applications will need to use a 'concrete subclass' of `XrSwapchainImageBaseHeader` instead (for instance, check `XrSwapchainImageBaseHeader.create(swapchainColorImages.address(), swapchainColorImages.capacity())` in the Vulkan example). I think we should also allow `XrSwapchainImageVulkan2KHR.Buffer`, `XrSwapchainImageVulkanKHR.Buffer`, `XrSwapchainImageOpenGLKHR.Buffer`, and `XrSwapchainImageOpenGLESKHR.Buffer`. I think the easiest way to implement this would be by simply adding a separate method for each struct (there are only 4 of them after all).

Spasi:
Hey @knokko,

> Make XrApiLayerProperties.type mutable. Currently, the example works around this problem using XRHelper.prepareApiLayerProperties, but I find it a bit silly that this work-around is needed.
> Do the same for XrExtensionProperties.

As I said in my previous reply, this has been done already (5e5cf08bb11567f98607c26f6ed7d0ce406c13f0). Will be available in the next snapshot (3.3.1 build 4).

Note that you'd still need to iterate over the structs in the buffer to fill in the type, so a utility method to do it might still be useful.

> Currently, xrEnumerateSwapchainImages only accepts XrSwapchainImageBaseHeader.Buffer as last parameter.

It would be nice to have overloads with the "concrete" types, in terms of API discoverability, but I think it would be a lot of Generator complexity for a few call sites. Another solution would be making `XrSwapchainImageOpenGLKHR.Buffer` extend `XrSwapchainImageBaseHeader.Buffer`, but it seems impossible with the current design.

How about adding "cast-to-supertype" methods? For example:

```java
XrSwapchainImageOpenGLKHR.Buffer swapchainImages = /* ... */;

xrEnumerateSwapchainImages(swapchain, imageCountOutput, swapchainImages.asXrSwapchainImageBaseHeader());
// or
xrEnumerateSwapchainImages(swapchain, imageCountOutput, swapchainImages.toXrSwapchainImageBaseHeader());
```

It would be a safe/convenient alternative to `XrSwapchainImageBaseHeader.create(swapchainImages.address(), swapchainImages.remaining())`.

knokko:
Whoops, I thought the type-next readonly patch was already released and these cases were somehow overlooked. Forget I said that.

I think adding overloads with the "concrete" types would be slightly nicer than using "cast-to-supertype" methods, but if this would be too much work, "cast-to-supertype" is also nice. But... why is adding overloads more difficult than adding "cast-to-supertype" methods? It sounds like both solutions would be a matter of hardcoding a few cases in the generator.

Spasi:
> But... why is adding overloads more difficult than adding "cast-to-supertype" methods? It sounds like both solutions would be a matter of hardcoding a few cases in the generator.

Adding "cast-to-supertype" methods means adding two simple methods for each struct type, one in the Struct subclass and one in the StructBuffer subclass.

Whereas adding overloads means adding them for every method that happens to accept the supertype. That is, potentially, a lot more methods generated. The implementation complexity is not comparable either, it would have to play nice with other Generator features. Also, what happens when a function accepts 2+ supertype parameters<sup>1</sup>? It could lead to a combinatorial explosion of overloads, especially when combined with Generator features that also create multiple overloads.

<sup>1</sup> - there is no such case now, but I have to consider what'll happen in the future and in other bindings, e.g. Vulkan.

knokko:
Ah... I thought you meant adding all hardcoded concrete overload cases manually. Doing it automatically could indeed cause some disturbance...

Ok, then I think adding `asSuperType` methods would be the best solution.

Spasi:
A way to recover discoverability is to inverse the cast. Instead of adding methods to the subtypes, we could overload the supertype's `create` method, for example:

```java
static XrSwapchainImageBaseHeader create(XrSwapchainImageOpenGLKHR value) { /* ... */ }
static XrSwapchainImageBaseHeader.Buffer create(XrSwapchainImageOpenGLKHR.Buffer value) { /* ... */ }
// and so on for other subtypes
```

and the call site becomes:

```java
xrEnumerateSwapchainImages(swapchain, imageCountOutput, XrSwapchainImageBaseHeader.create(swapchainImages));
```

What do you think?

knokko:
Discoverability is a bit nicer, but it might bloat the superclass if it has many subclasses. For instance, this would add 8 overloads to `XrSwapchainImageBaseHeader` (it has 4 subclasses, and each of them has its own buffer class). The `asSuperType` method would add only 2 overloads to each subclass. This adds the same number of method overloads in total, but they are more equally spread.

Spasi:
> but it might bloat the superclass if it has many subclasses

That's OK, as long as it scales linearly with the number of subtypes. We already do the same thing with `next`/`pNext` overloads (c26f72d7e24759412dd8aa2c245d936925e0553a).

Spasi:
The OpenXR bindings are in a great shape now and the above issues/suggestions have been addressed.

Closing this, thanks everyone for contributing!

PR #631, opened 2021-03-14
[WIP] Remove unsafe object scan.

mukel:
[WIP] Use JNI to compute field offsets instead of scanning the objects with Unsafe.

### DIFF
--- a/modules/lwjgl/core/src/main/c/org_lwjgl_system_MemoryUtil.c
+++ b/modules/lwjgl/core/src/main/c/org_lwjgl_system_MemoryUtil.c
@@ -14,4 +14,61 @@ JNIEXPORT jobject JNICALL Java_org_lwjgl_system_MemoryUtil_memGlobalRefToObject(
     return (jobject)(intptr_t)globalRef;
 }
 
+// static native long getInstanceFieldOffset(String className, String fieldName, String fieldDescriptor, Unsafe unsafeInstance) throws NoSuchFieldError;
+JNIEXPORT jlong JNICALL Java_org_lwjgl_system_MemoryUtil_getInstanceFieldOffset(JNIEnv *env, jclass clazz,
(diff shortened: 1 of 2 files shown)


httpdigest:
The reason for the _scanning_ is because the name of the Buffer's "address" field is not known (it can differ in different JREs, which don't use the OpenJDK class library).
Apart from that: Even if the field name _was_ known in advance, there wouldn't be a reason to use JNI, which in your case simply looks up the sun.misc.Unsafe's objectFieldOffset() method in native code and calling that method via JNI on the supplied Unsafe instance.
That could've just as well been implemented more directly in Java.
But, like I said: We cannot do this. We have to scan.

mukel:
Scanning objects like that is very fragile, field offsets are just a token/cookie, the physical location remains an implementation detail. The VM may decide the shuffle fields (packing), or even remove fields that are always constant or unused and provide a "fake" offset instead.
The issue was originally reported [here](https://github.com/oracle/graal/issues/3261). Espresso is an alternative JVM, we use the OpenJDK class library, but we use a different object model/layout which sometimes breaks spec-bending code.
Are you referring to the Android class library? Which alternative class library is incompatible?
Sorry for the sloppy workaround, I'd really like to give this a few more cycles, probably I can suggest an alternative, less invasive workaround to the scanning.

mukel:
Just read https://blog.lwjgl.org/memory-management-in-lwjgl-3/ 
To get the `DirectByteBuffer` address offset, would you accept rather scanning all fields to find it?
Peeking known fields in a type-safe manner to find a magic constant is an improvement (spec-compliant) over peeking with Unsafe.

Spasi:
Hey @mukel, I think I know how to implement this in a compliant way. Could you please close this PR and open an issue instead?

Thanks!

mukel:
Opened https://github.com/LWJGL/lwjgl3/issues/632

PR #640, opened 2021-03-28
build: update JOML to 1.10.1

Newbytee:


Spasi:
Hey @Newbytee, this was posted during a period in which I was inactive and I missed it when LWJGL work resumed. JOML is now up to `1.10.2` and `update-dependencies.xml` is already updated. Thanks though!

PR #644, opened 2021-04-10
fix APIUtil.apiGetManifestValue() for GraalVM native image build

chirontt:
The APIUtil.apiGetManifestValue() method doesn't work in a native image
generated by GraalVM, as ClassLoader.getResource() doesn't return a URL
starting with "jar:" as expected when running in standard JVM
environment. Instead, in native image which is run by an embedded
SubstrateVM, ClassLoader.getResource() returns a URL starting with
"resource:", for all resources in the classpath.

This fix would make APIUtil.apiGetManifestValue() work in both standard
JVM as well as in native image generated by GraalVM.

### DIFF
--- a/modules/lwjgl/core/src/main/java/org/lwjgl/system/APIUtil.java
+++ b/modules/lwjgl/core/src/main/java/org/lwjgl/system/APIUtil.java
@@ -19,6 +19,7 @@ import java.nio.file.*;
 import java.util.*;
 import java.util.function.*;
 import java.util.jar.*;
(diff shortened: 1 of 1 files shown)


Spasi:
Hey @chirontt, thanks for the comprehensive code!

I don't remember why we've been manually loading the manifest, instead of simply using `Package::getImplementationVersion`, but since `apiGetManifestValue` isn't used for something else anymore, I've decided to remove it. Your implementation will certainly be useful if we ever need more though.

PR #645, opened 2021-04-24
FreeBSD build (help needed)

ivan-volnov:
This is a try to build the library on FreeBSD x64

Carefully copied current linux build configuration, changed compiller to clang.

There is unresolved problem with openjfx14 files location. It seems lwjgl3 uses zulu openjdk build, but in FreeBSD we need to use openjdk8 + openjfx14

I have no idea how to fix it because I have no java experience at all.

Also created a PR https://github.com/LWJGL-CI/lwjgl3/pull/3 for CI Github Actions to test it.

If anybody wants to help, welcome!

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -216,6 +216,7 @@
             <src path="${core.kotlin.path}/libc"/>
             <src path="${core.kotlin.path}/libffi"/>
             <src path="${core.kotlin.path}/linux"/>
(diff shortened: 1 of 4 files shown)


Spasi:
Thank you @ivan-volnov, LWJGL now has FreeBSD support.

PR #651, opened 2021-05-21
Check for non-ASCII characters inside workaround check for JDK-8195129

### LINKED ISSUE #648: The check for JDK-8195129 can cause a library to be prematurely/duplicately loaded in a different `ClassLoader`
### Environment

* LWJGL version: `3.2.2` 
* LWJGL build #: `10`
* Java version: `1.8.0_292 AdoptOpenJDK`
* Platform: `Windows`
* Module: `core`

### Description

When in a complex classpath environment (such as the FabricMC `KnotClassLoader` which loads nested jar files), the check for JDK-8195129 causes the native library in question to be loaded during the "extraction" phase and before the "loading" phase. Crucially, the library isn't even loaded with the supplied class loader, it's loaded with the class loader of the `SharedLibraryLoader` class.

This has the effect of throwing an `UnsatisfiedLinkError` during the "loading" phase since the library has already been loaded.

The check in question: https://github.com/LWJGL/lwjgl3/blob/e4a6cc863f469ea8acfe3c2158f2c77d0c0aa95d/modules/lwjgl/core/src/main/java/org/lwjgl/system/SharedLibraryLoader.java#L279-L286

I propose ideally the check for JDK-8195129 should search the string for Unicode characters. If that isn't possible, then it should at least try to load the library using the class loader supplied to `SharedLibraryLoader::load`.

magneticflux-:
Closes #648

This change should maintain existing functionality of the JDK-8195129 check, but narrow it to only affect non-ASCII paths.

### DIFF
--- a/modules/lwjgl/core/src/main/java/org/lwjgl/system/SharedLibraryLoader.java
+++ b/modules/lwjgl/core/src/main/java/org/lwjgl/system/SharedLibraryLoader.java
@@ -315,7 +315,15 @@ final class SharedLibraryLoader {
     }
 
     private static boolean workaroundJDK8195129(Path file) {
-        return Platform.get() == Platform.WINDOWS && file.toString().endsWith(".dll");
+        boolean isWinDll = Platform.get() == Platform.WINDOWS && file.toString().endsWith(".dll");
+        if (isWinDll) {
+            // Check for unicode characters if this path is going to be loaded with System.load
+            for (char c : file.toString().toCharArray()) {
+                if (c > '\u007f') {
+                    return true;
+                }
+            }
+        }
+        return false;
     }
-
 }


magneticflux-:
@Spasi Could you take a look at this and check that it still works as in dedb44e0ab364ceab1f452daa825930052b1829c?

magneticflux-:
If you're familiar with mixins (a runtime bytecode manipulation framework), this is what I used to confirm the fix in the full application that encountered this issue:

```java
@Mixin(targets = "org.lwjgl.system.SharedLibraryLoader", remap = false)
public class SharedLibraryLoaderMixin {
    @Redirect(method = "canWrite", at = @At(value = "INVOKE", target = "Ljava/lang/System;load(Ljava/lang/String;)V"))
    private static void redirectLoad(String filename) {
        if (!CharMatcher.ascii().matchesAllOf(filename)) {
            throw new RuntimeException();
        }
    }
}
```

Spasi:
Thanks @magneticflux-! The check has been added manually with c5bab97e18e70d90a5e5f033f5af7ccf81cb9fbd (slightly changed to avoid the `char[]` allocation).

PR #713, opened 2022-01-03
Fix SharedLibraryUtil#getLibraryPath returning NUL

octylFractal:
The length includes the null-terminator, but memUTF8 assumes there isn't one.


Spasi:
Thank you @octylFractal!

PR #716, opened 2022-01-11
fix(jemalloc): update jemalloc version in bindings to 5.2.1

mrmangohands:
It seems this was checked off in b8aaf7355f78f4ffc12667607203bbefa9fbaca3, but the actual change never made it in. As a result any uses of memoryutil on windows suffer from the nasty leak fixed in jemalloc 5.2.1 with current releases, but bumping it should resolve things.

### DIFF
--- a/modules/lwjgl/jemalloc/src/generated/java/org/lwjgl/system/jemalloc/JEmalloc.java
+++ b/modules/lwjgl/jemalloc/src/generated/java/org/lwjgl/system/jemalloc/JEmalloc.java
@@ -74,13 +74,13 @@ public class JEmalloc {
     public static final int JEMALLOC_VERSION_MINOR = 2;
 
     /** The bugfix version. */
-    public static final int JEMALLOC_VERSION_BUGFIX = 0;
+    public static final int JEMALLOC_VERSION_BUGFIX = 1;
 
     /** The revision number. */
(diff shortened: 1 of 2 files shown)


Spasi:
Hey @mrmangohands,

Thanks for the patch, `JEMALLOC_VERSION_BUGFIX` and `JEMALLOC_VERSION_GID` in the jemalloc bindings have the wrong values indeed. However, this has nothing to do with the jemalloc binary shipped with LWJGL, which is actually the correct 5.2.1 build. You can verify this by inspecting `jemalloc.dll.git` in `lwjgl-jemalloc-natives-windows.jar`.

Have you actually been able to reproduce a memory leak?

Spasi:
Btw, if you'd like to have a look, the jemalloc binaries are built at [LWJGL-CI/jemalloc](https://github.com/LWJGL-CI/jemalloc) using [Github Actions](https://github.com/LWJGL-CI/jemalloc/blob/master/.github/workflows/lwjgl.yml).

mrmangohands:
Ah, my mistake, a minecraft mod called Sodium ran into this issue in the middle of last year and I was misremembering what version of lwjgl minecraft uses, it's still on lwjgl 3.2.2 for windows which explains things

mrmangohands:
Removed the mentions in the notes file and updated the commit message

Spasi:
Thanks!

PR #717, opened 2022-01-12
feat(OpenXR) construct projection matrix via JOML built-ins

httpdigest:
Since we are given the angles between the left, right, bottom
and top frustum planes, respectively, and the line
perpendicular to the near/far frustum plane, we can use
glFrustum (i.e. Matrix4f.frustum) directly, providing it with
the distances to the frustum planes (via Math.tan(angle)).


httpdigest:
In addition, JOML 1.10.4 will feature [`[set]perspectiveOffCenterFov()`](https://github.com/JOML-CI/JOML/commit/21db7f06b271020611f44d66cbc49bd12eed2f9d), which will just do the Math.tan() internally.

httpdigest:
I, of course, checked that the proposed changes:
1. compile
2. produce exactly the same results (effective projection matrix) compared to the previous method in the XRHelper class

And in addition, the round-trip between NIO buffer -> matrix is not needed anymore.

PR #725, opened 2022-02-02
fix(OpenXR): Invert y-axis of projection matrix in Vulkan example

knokko:
While refactoring `XRHelper.applyProjectionToMatrix`, we forgot to account for the fact that the y-axis is 'upside-down' in the Vulkan coordinate system. So currently, the Vulkan example is basically upside down (this looks really weird when rotating the head vertically since the projection matrix basically 'counters' it).

This is probably the easiest solution to account for the 'upside-down' y-axis. The OpenXR + OpenGL demo looks good.


httpdigest:
Yeah, inverting clip-space Y is usually how it should be done.
Other options are using KHR_maintenant1 (core in VK 1.1) and use negative viewport height.
See: https://www.saschawillems.de/blog/2019/03/29/flipping-the-vulkan-viewport/

knokko:
I prefer my solution over the negative viewport solution because it is easier and doesn't require extensions (even if the extension is extremely common, it is still extra effort).

Spasi:
Thanks!

PR #726, opened 2022-02-02
OpenXR demo improvements

knokko:
This merge request fixes some critical problems in the OpenXR demos (I wonder why I never noticed some of these before...), and it adds partial Linux support to the OpenXR + OpenGL demo. It is partial because it only supports the X11 windowing system at the moment.


Spasi:
Thanks!

PR #743, opened 2022-03-03
fix VK_API_VERSION_PATCH

Kantaja:
the implementation of VK_API_VERSION_PATCH added in c4815d3 mistakenly returns bits 22-31 instead of 0-11


Spasi:
@Kantaja Thanks!

Issue #554, opened 2020-05-02
`apply` functions added in 3.2.3 break existing Kotlin code


Hugobros3:
Release 3.2.3 brought a new `apply` method in `StructBuffer` and thus clashes with the built-in `apply` extension function available in Kotlin code. That apply function appears to do the same job but has a different calling convention from the Kotlin one (Kotlin `apply` takes a lambda with `this` renamed to the object it's called on, while `StructBuffer`'s takes a SAM ).

The result of this is large portions of Kotlin code that use `StructBuffer`s are now broken (example [here](https://cdn.discordapp.com/attachments/600646864907403264/706216347704754317/unknown.png)) and need tedious manual working around the issue. I have to say it's a very unfortunate oversight to have implemented any functionality that clashes with basic Kotlin stdlib functionality, especially given LWJGL itself uses Kotlin for various things internally.

I'm not sure how to fix this, I admit it's been a while since I worked on my Kotlin game engine project. I'd say roll back the changes, but this already shipped so it's a matter of choosing whose code you'll break now. The middle ground might just be to rename the function ?

Hugobros3:
(I don't know about *tons*, it's just mine afaik. But `apply` is a stable of idiomatic Kt code)

octylFractal:
I don't think this does the same job -- here `apply` works on a relative element of the buffer (and increments the position!), whereas Kotlin's `apply` would be targeting the buffer itself. Your code only worked because the buffer also has methods to target the currently pointed-to element of the buffer, not because Kotlin's `apply` was extracting the element from the buffer.

I do think that it may be slightly misleading for them to have the same name and similar signatures, but do _very_ different things.

XenoAmess:
So can it build with the latest version of kotlin?
Or just compile error?

Hugobros3:
Your question is weirdly phrased. This has nothing to do with changes on the Kotlin side, the latest version of LWJGL just breaks my code in plenty of places. 

I do very liberal amounts of stackAlloc(1) to get singleton array of structs to pass into Vulkan in plenty of places where the API expects an array, since LWJGL doesn't offer overloads for the non-buffer variants. Using Kotlin's `apply` there just makes sense and yields the most readable code, which is why I had settled on that pattern.

XenoAmess:
> Your question is weirdly phrased. This has nothing to do with changes on the Kotlin side, the latest version of LWJGL just breaks my code in plenty of places.
> 
[...]

PR #745, opened 2022-03-04
feat(core) add StructBuffer.applyEach(IndexedConsumer)

httpdigest:
Often, when creating `Struct.Buffer` instances, the values of particular fields of each struct within the buffer is sourced from another indexed data structure, like an array or list at position `i`.

Currently, in order to access those other data structures, we have to use an external iteration (such as an indexed for-loop or a for-each loop driven by the other data structure) and use either the relative `apply(Consumer)` or absolute `apply(index, Consumer)` calls on `StructBuffer` within that loop.

It'd be nice to compose whole structs and struct buffers without external iteration, only relying on internal iterators, like those provided by `StructBuffer.forEach()`. However, the downside of `StructBuffer.forEach()` is that it does not return `this` and it _also_ does not provide the iterator index in order to access another data structure for the currently iterated struct.

This change proposes to add a mix of `StructBuffer.apply()` and `StructBuffer.forEach()` to achieve both an internal iteration design and allowing to return `this` StructBuffer.


TheMrMilchmann:
Looks useful but I have two minor points: 

1. Maybe we should ~~steal~~ borrow from Kotlin here and call it [`onEachIndexed`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/on-each-indexed.html) (or `applyEachIndexed` to increase discoverability).
2. Instead of adding a custom interface, [`ObjIntConsumer`](https://docs.oracle.com/javase/8/docs/api/java/util/function/ObjIntConsumer.html) could be used. However, note that the argument order is reversed in that interface which is somewhat unconventional. (I assume `BiConsumer` was not used to avoid boxing.)

httpdigest:
Thanks, I agree with both suggestions.

Spasi:
Won't this cause trouble similar to #554?

httpdigest:
I read that issues as the particular extension function `apply` being a problem. Not any methods named differently, like `applyEachIndexed`. I can't find any source that reference `applyEachIndexed` with Kotlin.

PR #771, opened 2022-06-13
Try to extract build suffix from JAR file

### LINKED ISSUE #770: Version reports `SNAPSHOT` on module-path
### Version

3.3.1

### Platform

Linux x64, Linux arm64, Linux arm32, macOS x64, macOS arm64, Windows x64, Windows x86, Windows arm64

### JDK

JDK 9+

### Module

LWJGL Core

### Bug description

**Actual**

`java --module-path lib --add-modules org.lwjgl org.lwjgl.Version`
```
3.3.1 SNAPSHOT
3.3.1
```

**Expected**

`java --class-path lib/org.lwjgl.jar org.lwjgl.Version`
```
3.3.1 build 7
3.3.1
```

This is due to the following specification in [java.lang.Package](https://docs.oracle.com/en/java/javase/18/docs/api/java.base/java/lang/Package.html)

> A Package automatically defined for classes in a **named module** has the following properties:
>
> - The name of the package is derived from the [binary names](https://docs.oracle.com/en/java/javase/18/docs/api/java.base/java/lang/Class.html#getName()) of the classes. Since classes in a named module must be in a named package, the derived name is never empty.
> - The package is sealed with the [module location](https://docs.oracle.com/en/java/javase/18/docs/api/java.base/java/lang/module/ModuleReference.html#location()) as the code source, if known.
> - **The specification and implementation titles, versions, and vendors are unspecified.** ⬅️
> - Any annotations on the package are read from `package-info.class` as specified above.

Therefore `Version.class.getPackage().getImplementationVersion()` as [used here](https://github.com/LWJGL/lwjgl3/blob/ad602750ad87939276d1889b28a757061e90715b/modules/lwjgl/core/src/main/java/org/lwjgl/Version.java#L25) returns `null`.

### Notes

Note that the compiled module descriptors do contain the correct version information.

`java --module-path lib --list-modules`
```
...
org.lwjgl@3.3.1+7 file:///.../org.lwjgl.jar
org.lwjgl.glfw@3.3.1+7 file:///.../org.lwjgl.glfw.jar
org.lwjgl.glfw.natives@3.3.1+7 file:///.../org.lwjgl.glfw.natives.jar
org.lwjgl.natives@3.3.1+7 file:///.../org.lwjgl.natives.jar
org.lwjgl.opengl@3.3.1+7 file:///.../org.lwjgl.opengl.jar
org.lwjgl.opengl.natives@3.3.1+7 file:///.../org.lwjgl.opengl.natives.jar
```

sormuras:
Fixes #770

### DIFF
--- a/modules/lwjgl/core/src/main/java/org/lwjgl/Version.java
+++ b/modules/lwjgl/core/src/main/java/org/lwjgl/Version.java
@@ -4,7 +4,11 @@
  */
 package org.lwjgl;
 
+import javax.annotation.*;
+import java.io.*;
+import java.net.*;
 import java.util.*;
+import java.util.jar.*;
 
 /** This class can be used to query the LWJGL version. */
 public final class Version {
@@ -18,18 +22,19 @@ public final class Version {
     /** The development state of the current build. */
     public static final BuildType BUILD_TYPE = BuildType.STABLE;
 
-    private static final String version =
+    private static final String versionPlain =
         String.valueOf(VERSION_MAJOR) +
         '.' + VERSION_MINOR +
-        '.' + VERSION_REVISION + BUILD_TYPE.postfix +
-        ' ' + Optional.ofNullable(Version.class.getPackage().getImplementationVersion()).orElse("SNAPSHOT");
+        '.' + VERSION_REVISION + BUILD_TYPE.postfix;
+
+    private static final String version = versionPlain + VersionImpl.find();
 
     private Version() {
     }
 
     public static void main(String[] args) {
         System.out.println(version);
-        System.err.println(version.substring(0, version.indexOf(' ')));
+        System.err.println(versionPlain);
     }
 
     /** Returns the LWJGL version. */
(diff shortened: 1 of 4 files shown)


modules/lwjgl/core/src/main/java/org/lwjgl/Version.java
```diff
@@ -53,4 +55,15 @@ public enum BuildType {
...
+    private static Optional<String> optionalBuildSuffix() {
+        String suffix = Version.class.getPackage().getImplementationVersion();
+        if (suffix != null) return Optional.of(suffix);
+        try {
+          try (JarFile jar = new JarFile(Path.of(Version.class.getProtectionDomain().getCodeSource().getLocation().toURI()).toFile())) {
+            return Optional.ofNullable(jar.getManifest().getMainAttributes().getValue(Attributes.Name.IMPLEMENTATION_VERSION));
+          }
+        } catch (Exception ignored) {
+            return Optional.empty();
+        }
```
TheMrMilchmann, line 67: Instead of loading the JarFile manually, I suggest just loading the `Manifest` manually by retrieving the resource via `Class.getResourceAsStream(String)`. This should work on Java 8 and Java 9+ and even with modules, since "a resource name with the name "META-INF/MANIFEST.MF" is never encapsulated because "META-INF" is not a legal package name." (See the [encapsulation rules for resources](https://docs.oracle.com/en/java/javase/18/docs/api/java.base/java/lang/Module.html#getResourceAsStream(java.lang.String))). A few basic tests seem to confirm that this works as expected.

```suggestion
        try (InputStream in = Version.class.getResourceAsStream("/" + JarFile.MANIFEST_NAME)) {
            if (in == null) return Optional.empty();

            Manifest manifest = new Manifest(in);
            return Optional.ofNullable(manifest.getMainAttributes().getValue(Attributes.Name.IMPLEMENTATION_VERSION));
        } catch (Exception ignored) {
            return Optional.empty();
        }
```

TheMrMilchmann (changes requested):

modules/lwjgl/core/src/main/java/org/lwjgl/Version.java
sormuras: Check via:
- `jshell --module-path .lib --add-modules ALL-MODULE-PATH`
- `jshell> org.lwjgl.Version.class.getResource("/" + java.util.jar.JarFile.MANIFEST_NAME)`
- `$1 ==> jar:file:///.../lib/org.lwjgl.jar!/META-INF/MANIFEST.MF`

sormuras: Seems to work and might also spare some memory/OS resources by re-using already loaded bits.

TheMrMilchmann (approved):

Spasi:
Thank you @sormuras!

Issue #772, opened 2022-06-13
HelloVulkan validation errors


TheChocolateOre:
### Version

3.3.1

### Platform

Windows x64

### JDK

OpenJDK 18.0.1

### Module

Vulkan

### Bug description

With validation layers enabled, the HelloVulkan.java runs and produces the expected results (a window with a triangle). But, I get a few errors, which don't crash the app, but still might be of concern.

Vulkan version: 0.1.3.204 (Intel HD Graphics 530)
Vulkan SDK: 1.3.211.0 (latest as of writing)

### Stacktrace or crash log output

```shell
ERROR: [Validation] Code 0 : Validation Error: [ VUID-vkCmdPipelineBarrier-dstAccessMask-02816 ] Object 0: handle = 0x1a289fd77a8, type = VK_OBJECT_TYPE_COMMAND_BUFFER; | MessageID = 0x69c8467d | vkCmdPipelineBarrier(): .pImageMemoryBarriers[0].dstAccessMask bit VK_ACCESS_DEPTH_STENCIL_ATTACHMENT_WRITE_BIT is not supported by stage mask (VK_PIPELINE_STAGE_TOP_OF_PIPE_BIT). The Vulkan spec states: The dstAccessMask member of each element of pMemoryBarriers must only include access flags that are supported by one or more of the pipeline stages in dstStageMask, as specified in the table of supported access types (https://vulkan.lunarg.com/doc/view/1.3.211.0/windows/1.3-extensions/vkspec.html#VUID-vkCmdPipelineBarrier-dstAccessMask-02816)
[...]

PR #777, opened 2022-07-08
Fix validation errors in the vulkan demo

knokko:
As reported in https://github.com/LWJGL/lwjgl3/issues/772, running the current `HelloVulkan` demo causes validation errors. This pull request fixes these validation errors.


Spasi:
Thanks!

PR #799, opened 2022-08-19
Support JNI_VERSION_19 fixing crash with Java 19 EA when preview features are enabled.

modmuss50:
[JEP 425](https://openjdk.org/jeps/425) (Project Loom) added the JNI function `IsVirtualThread`. This PR updates `ThreadLocalUtil` to support the addion of this new function and JNI version.

JDK change: https://github.com/openjdk/jdk/commit/53bf1bfdabb79b37afedd09051d057f9eea620f2
JDK issue: https://bugs.openjdk.org/browse/JDK-8286176

This change fixes a crash when running on the latest Java 19 EA build with preview features enabled. This was noticed when testing with Minecraft 1.19.2 on Windows + Java 19 EA with preview features enabled.

Please let me know if any further information or testing is required 👍

### DIFF
--- a/doc/notes/3.3.2.md
+++ b/doc/notes/3.3.2.md
@@ -25,6 +25,7 @@ This build includes the following changes:
 #### Improvements
 
 - Windows: Shared libraries are now built with Visual Studio 2022 (up from 2019)
+- Core: Added support for JDK 19 (#799)
 - Core: Added `MemoryUtil::memByteBuffer(Struct)`. It creates a `ByteBuffer` view of a struct value.
 - Core: More `SharedLibraryLoader` improvements. (#790)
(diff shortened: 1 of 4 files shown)


Spasi:
Thank you @modmuss50!

gudenau:
@Spasi Would it be possible to change the implementation to scan for a null pointer at the end of the structure? Unless the structure grows to be on an allocation border I would think that scanning would be more robust.

Spasi:
@gudenau Hey, I don't think that would be a reliable fix.

In any case, new JNI functions are rare, it's only happened twice in LWJGL3's lifetime (Java 9 and Java 19). I'll add a `Configuration` override, if it happens again, a workaround without updating LWJGL will be available.

PR #819, opened 2022-10-13
chore: use SPDX license identifier in POM

TheMrMilchmann:
Simply change the license `name` in the POM to a [valid SPDX identifier](https://spdx.org/licenses/BSD-3-Clause.html). This allows tools to automatically detect the license for LWJGL artifacts and is even [recommended in the POM documentation](https://maven.apache.org/pom.html#licenses).

For consistency, I updated the dummy POM too.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -360,7 +360,7 @@ publishing {
 
             licenses {
                 license {
-                    name.set("BSD")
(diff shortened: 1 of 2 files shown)


Spasi:
@TheMrMilchmann thanks!

PR #825, opened 2022-11-13
fix: prevent debug message tearing in multithreaded environments

TheMrMilchmann:
There is a tiny chance that the split PrintStream writes could tear in multithreaded environments when writing a lot of messages concurrently. This change uses monitors to prevent such issues going forward.

Since these paths are not performance sensitive, the use of monitors shouldn't be an issue here. Alternatively, we could prebuild and log a single `String` but I think that would cause undesirable output for multiline strings.

### DIFF
--- a/modules/lwjgl/core/src/main/java/org/lwjgl/system/APIUtil.java
+++ b/modules/lwjgl/core/src/main/java/org/lwjgl/system/APIUtil.java
@@ -80,10 +80,20 @@ public final class APIUtil {
      *
      * @param msg the message to print
      */
-    public static void apiLog(@Nullable CharSequence msg) {
+    public static void apiLog(CharSequence msg) {
         if (DEBUG) {
-            DEBUG_STREAM.print("[LWJGL] ");
-            DEBUG_STREAM.println(msg);
+            DEBUG_STREAM.print("[LWJGL] " + msg + "\n");
+        }
+    }
+
+    /**
+     * Same as {@link #apiLog}, but replaces the LWJGL prefix with a tab character.
+     *
+     * @param msg the message to print, in continuation of a previous message
+     */
+    public static void apiLogMore(CharSequence msg) {
+        if (DEBUG) {
+            DEBUG_STREAM.print("\t" + msg + "\n");
         }
     }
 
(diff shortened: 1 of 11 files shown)


modules/lwjgl/core/src/main/java/org/lwjgl/system/APIUtil.java
```diff
@@ -80,10 +80,9 @@ private APIUtil() {
      *
      * @param msg the message to print
      */
-    public static void apiLog(@Nullable CharSequence msg) {
+    public static void apiLog(CharSequence msg) {
```
TheMrMilchmann, line 83: Not sure why I marked this as `@Nullable` in the first place. While the method could technically accept `null`, we probably don't ever want to do that.

Spasi:
Hey @TheMrMilchmann,

I don't think message tearing is that big a deal, but sure, it should be avoided if possible. Not with `synchronized` though, especially after the removal of biased locking in JDK 15. Note that `PrintStream` also internally synchronizes on every message printed.

I would be more comfortable with a solution that buffers up text in a `StringBuilder` and does a single `print`. Could you please refactor this PR to do that? (in the debug allocator, please do a print per `Allocation`, not a single print for everything)

TheMrMilchmann:
> I don't think message tearing is that big a deal, but sure, it should be avoided if possible.

I fully agree but figured that it's a fairly small change. There are a few more writes that could tear (some call sites of `apiLog` in `LibraryResource` and `SharedLibraryLoader`) but those are not as easy to fix and the chance for that to occur is even rarer so I didn't bother to touch these.

> Not with `synchronized` though, especially after the removal of biased locking in JDK 15. Note that `PrintStream` also internally synchronizes on every message printed.

I assumed that we don't care about the cost of monitors here because none of these sites should be hot paths. The only exception I can think of might be the different GL debug extensions depending on how verbose the implementations are. But that should usually only affect debugging environments.

> I would be more comfortable with a solution that buffers up text in a `StringBuilder` and does a single `print`. Could you please refactor this PR to do that? (in the debug allocator, please do a print per `Allocation`, not a single print for everything)

Sure, I changed it accordingly. I added an additional commit for now and can squash later, if you're happy with the change. Additionally, I noticed that I forgot to include the change to the template for the GLFW callback. That's fixed now too.

PR #844, opened 2023-01-08
Upgrade Gradle 6.6 -> 7.6

tonihele:
Upgrades the Gradle wrapper to 7.6. Also added the version to the build file wrapper task. At least the build LOCAL works. Doesn't produce any warnings either.


Spasi:
Thanks @tonihele!

I updated Gradle to `8.0.2` (released yesterday). Also removed the version from the build file, I prefer to pass it explicitly via the cli (e.g. `gradle wrapper --gradle-version 8.0.2 --distribution-type bin`).

PR #851, opened 2023-01-21
Crypt32 bindings

TheMrMilchmann:
Bindings for [dpapi.h and crypt32.dll](https://learn.microsoft.com/en-us/windows/win32/api/dpapi/) providing support for:

- `CryptProtectData` and `CryptUnprotectData`, and
- `CryptProtectMemory` and `CryptUnprotectMemory`.

This is not fully finished yet and I'll add some questions inline.


modules/lwjgl/core/src/templates/kotlin/core/windows/templates/Crypt32.kt
```diff
@@ -0,0 +1,175 @@
...
+
+    SaveLastError..BOOL(
+        "CryptProtectMemory",
+        """
+        The {@code CryptProtectMemory} function encrypts memory to prevent others from viewing sensitive information in your process. For example, use the
+        {@code CryptProtectMemory} function to encrypt memory that contains a password. Encrypting the password prevents others from viewing it when the process
+        is paged out to the swap file. Otherwise, the password is in plaintext and viewable by others.
+        """,
+
+        void.p("pDataIn", "a pointer to the block of memory to encrypt"),
```
TheMrMilchmann, line 94: `pDataIn` is actually defined as `LPVOID` ([ref](https://learn.microsoft.com/en-us/windows/win32/api/dpapi/nf-dpapi-cryptprotectmemory)). However, this is mapped to a raw pointer instead of a buffer and does not also not allow us to autosize the `cbDataIn` parameter. This, I chose `void.p` instead. Is this fine, or should I change this to align with the definition?

modules/lwjgl/core/src/templates/kotlin/core/windows/templates/Crypt32.kt
```diff
@@ -0,0 +1,175 @@
...
+            free its {@code pbData} member by calling the #LocalFree() function.
+            """
+        )
+    )
+
+    SaveLastError..BOOL(
+        "CryptUnprotectMemory",
+        "The {@code CryptUnprotectMemory} function decrypts memory that was encrypted using the #CryptProtectMemory() function.",
+
+        void.p("pDataIn", "a pointer to the block of memory to decrypt"),
```
TheMrMilchmann, line 160: See `CryptProtectMemory`

modules/lwjgl/core/src/templates/kotlin/core/windows/WindowsTypes.kt
```diff
@@ -631,4 +633,28 @@ val INPUT = struct(Module.CORE_WINDOWS, "INPUT") {
...
+}
+
+val CRYPTPROTECT_PROMPTSTRUCT = struct(Module.CORE_WINDOWS, "CRYPTPROTECT_PROMPTSTRUCT") {
+    documentation =
+        """
+        The {@code CRYPTPROTECT_PROMPTSTRUCT} structure provides the text of a prompt and information about when and where that prompt is to be displayed when
+        using the #CryptProtectData() and #CryptUnprotectData() functions.
+        """
+
+    DWORD("cbSize", "the size, in bytes, of this structure")
```
TheMrMilchmann, line 656: Can we autosize this somehow?

modules/lwjgl/core/src/generated/java/org/lwjgl/system/windows/Crypt32.java
```diff
@@ -0,0 +1,230 @@
...
+
+import java.nio.*;
+
+import org.lwjgl.*;
+
+import org.lwjgl.system.*;
+
+import static org.lwjgl.system.APIUtil.*;
+import static org.lwjgl.system.MemoryStack.*;
+import static org.lwjgl.system.MemoryUtil.*;
```
TheMrMilchmann, line 18: This is missing an import for `org.lwjgl.system.Checks.*;` which seems to be a bug in the generator.

I assume this is caused by a discrepancy between 

https://github.com/LWJGL/lwjgl3/blob/ee2dc3bbdc8d601428c1108cf8486e2aaf18fd07/modules/generator/src/main/kotlin/org/lwjgl/generator/NativeClass.kt#L472-L485 and https://github.com/LWJGL/lwjgl3/blob/ee2dc3bbdc8d601428c1108cf8486e2aaf18fd07/modules/generator/src/main/kotlin/org/lwjgl/generator/Functions.kt#L471-L477 but I'm not sure how to best solve this.

TheMrMilchmann (commented):
Added a couple of comments inline.

TheMrMilchmann:
(Pinging @Spasi to have a look at this because I'm not sure if GitHub notifies people about draft PRs. No hurry though. 😅)

modules/lwjgl/core/src/templates/kotlin/core/windows/templates/Crypt32.kt
Spasi: Changed to `typedef(void.p, "LPVOID")("pDataIn", "a pointer to the block of memory to encrypt")`. Functionally the same, but changes the annotation to `@NativeType("LPVOID")`.

modules/lwjgl/core/src/templates/kotlin/core/windows/WindowsTypes.kt
Spasi: Yeah, with `Expression("SIZEOF")..`. Adds the `cbSize$Default()` overload, similar to the Vulkan bindings.

modules/lwjgl/core/src/generated/java/org/lwjgl/system/windows/Crypt32.java
Spasi: Thanks, fixed!

Spasi:
Hey @TheMrMilchmann,

I've added a commit that addresses the above questions. Will split/squash/cleanup the commits as necessary when it's time to merge this.

TheMrMilchmann:
Thanks, I just looked at your changes and I believe you have already taken care of all remaining issues. In my opinion, this is good to go. However, I'm not sure where to mention this in the changelog. Could you also take care of this? Otherwise, let me know if anything else is missing.

Spasi:
Thanks @TheMrMilchmann!

Did one more change before the merge, added `IgnoreMissing..` to `CryptProtectMemory` and `CryptUnprotectMemory` because technically we support Windows XP and these two are available on Vista+.

PR #927, opened 2023-09-29
Fix memory leak in Callback.free

kb-1000:
CallbackRegistry.remove appears to be unused currently, meaning creating and freeing callbacks on platforms that *don't* use the no-op registry leaks memory.


Spasi:
Thank you @kb-1000!

PR #936, opened 2023-10-27
LLVM target initialization bindings for AArch64, ARM, AVR, Mips, PPC & WASM

KitsuneAlex:
As the title says, this PR adds target initialization bindings for some major CPU architectures which are missing in the LLVM module at the moment. Cheers.


KitsuneAlex:
@Spasi would be nice if you could take a look at this :)

Spasi:
Thank you @KitsuneAlex!

PR #979, opened 2024-05-05
fix(Nuklear): wrong autogenerated point count

### LINKED ISSUE #978: wrong point count for nk_fill_polygon, nk_stroke_polygon, and nk_stroke_polyline
### Version

3.3.3

### Platform

Linux x64

### JDK

openjdk 17.0.11 2024-04-16

### Module

Nuklear

### Bug description

I suspect that the array size instead of the number of points (= array size divided by two) is passed to Nuklear.
Here is a Clojure snippet using a float buffer of size 6.

image

 Instead of rendering a blue triangle and a white polyline, the polygon seems to be connected to the point `(0, 0)`:

image

The `nk_fill_polygon` and similar methods need to be [passed a point count](https://github.com/KenthJohan/eg/blob/ab996a64f2430b22efd2484bed0b5d5abd069068/apps/demo_nuklear/src/canvas.c#L75).

This issue also has been reported [here](https://github.com/vurtun/nuklear/issues/808).

### Stacktrace or crash log output

_No response_

wedesoft:
The autogenerated point count for polygon operations is wrong (array size needs to be divided by two).
I checked that the following methods are affected:
* stroke_polyline
* stroke_polygon
* fill_polygon

Note that I didn't manage to build new JAR files, so I was not able to test the fix.

See below issue for screenshots.

Closes #978

### DIFF
--- a/doc/notes/3.3.4.md
+++ b/doc/notes/3.3.4.md
@@ -44,6 +44,7 @@ This build includes the following changes:
 - Core: Fixed callback wrapper memory leak with the CHM closure registry. (#927)
 - LLVM: Fixed `LLVMGetBufferStart` to return `ByteBuffer` instead of `String`. (#934)
 - LLVM: Fixed `LookupIntrinsicID` to return `unsigned` instead of `void`. (#950)
+- Nuklear: Fixed auto-sizing of `nk_stroke_polyline`, `nk_stroke_polygon`, `nk_fill_polygon` buffers. (#978)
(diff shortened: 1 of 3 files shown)


wedesoft:
Ok, managed to create updated JAR files. The fix works:
fixpoly

Spasi:
Thank you @wedesoft!

wedesoft:
Thanks for maintaining LWJGL3 :)

PR #993, opened 2024-07-08
fix(bgfx) Change parameters for `is_frame_buffer_valid`

bonenaut7:
BGFX's method `bgfx::isFrameBufferValid` takes array size, and array reference of `Attachment`'s.
However in the bindings there is only single `BGFXAttachment` object, instead of `BGFXAttachment.Buffer` as it's done in the `bgfx_create_frame_buffer_from_attachment`.

I'm replaced `MapToInt` with `AutoSize`, now `is_frame_buffer_valid` has `BGFXAttachment.Buffer` and same buffer could be utilized both for validating attachments and creating framebuffer from attachments.


Spasi:
Thank you @bonenaut7!

PR #1002, opened 2024-08-16
Add support for glyph index based msdfgen-ext APIs

KitsuneAlex:
This PR adds support for the following msdfgen-ext functions:
- `msdfgen::loadGlyph(msdfgen::Shape, msdfgen::FontHandle, msdfgen::GlyphIndex, int)` for loading glyphs by their glyph ID directly instead of their codepoint
- `msdfgen::getGlyphIndex(msdfgen::FontHandle, unsigned)` for retrieving the glyph index of a given codepoint
- `msdfgen::getKerning(double&, msdfgen::FontHandle, unsigned, unsigned)` for retrieving the kerning between the two given unicode codepoints
- `msdfgen::getKerning(double&, msdfgen::FontHandle, msdfgen::GlyphIndex, msdfgen::GlyphIndex)` for retrieving the kerning between the two given glyphs referenced by their indices

This allows using this library in conjunction with other libraries like harfbuzz.

### DIFF
--- a/doc/notes/3.3.5.md
+++ b/doc/notes/3.3.5.md
@@ -25,6 +25,8 @@ This build includes the following changes:
 
 #### Improvements
 
+- msdfgen: Added support for glyph index based msdfgen-ext APIs. (#1002)
+
 #### Fixes
 
 - Core: Fixed support for `va_list` parameters.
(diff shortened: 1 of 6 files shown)


Spasi:
Thank you @KitsuneAlex!

Issue #875, opened 2023-04-08
How does ThreadLocalUtil#setCapabilities works?


Berstanio:
### Question

Hey,

first of all I hope this is the right place to ask such question.
I'm currently experementing with lwjgl3 and native-image. Currently I get a segault in ThreadLocalUtil#setCapabilities and I would like to understand the method first, before I proceed with e.g. opening a graal issue.

```java
    public static void setCapabilities(long capabilities) {
        // Get thread's JNIEnv
        long env_pp = getThreadJNIEnv();
        long env_p  = memGetAddress(env_pp);

        if (capabilities == NULL) {
            if (env_p != JNI_NATIVE_INTERFACE) {
                memPutAddress(env_p + CAPABILITIES_OFFSET,
                    // FUNCTION_MISSING_ABORT table
                    memGetAddress(JNI_NATIVE_INTERFACE + CAPABILITIES_OFFSET));
            }
        } else {
            if (env_p == JNI_NATIVE_INTERFACE) {
                setupEnvData();
                env_p = memGetAddress(env_pp);
            }
            memPutAddress(env_p + CAPABILITIES_OFFSET, capabilities);
        }
    }
```

So, how I understand it `capabilities` a array of function pointer. This array will be later used to dispatch different GL functions.
```c
struct JNINativeInterface {
    void*       reserved0;
    void*       reserved1;
    void*       reserved2;
    void*       reserved3;
[...]
```
Then, in the line `memPutAddress(env_p + CAPABILITIES_OFFSET, capabilities);`, the function array will be put inside `reserved3` field in the JNIEnv struct.
From there it will be used to dispatch the functions.

Is that correct?

Berstanio:
Whoooops, I didn't realised that all the code comments are on the top of the class... I'm sorry!
```
    - On setCapabilities:
        * If necessary, a jniNativeInterface copy is created and injected to the current thread (JavaThread::_jni_environment points to the copy).
        * A pointer to the capabilities function pointer array is set to jniNativeInterface::reserved3.
```

So yeah, it works as I expected it to do. 

Now I just have a follow up question, do you know, whether there is somewhere any documentation to the "reserved3" field in the struct? I couldn't find any.

Spasi:
Hey @Berstanio,

There's no documentation, it's just some reserved space before the function pointers. It was never used for anything and `reserved0-3` are still unused up to the latest OpenJDK commit at the time I'm writing this.

[...]

PR #1003, opened 2024-08-19
feat(core) EspressoVM compatibility

lewurm:
`reserved0` is used by EspressoVM, thus it cannot serve as a "reference NULL" as on HotSpot: https://github.com/oracle/graal/blob/3149f62458029ffe92b8dcefe0b3e59612684cfa/espresso/src/com.oracle.truffle.espresso.mokapot/include/mokapot.h#L67-L73

The relevant config is `MOKA_LATTE`. `reserved3` is `NULL` though.

-----

EspressoVM aka. Java on Truffle: https://www.graalvm.org/latest/reference-manual/java-on-truffle/

I tested the fix with Espresso, but I'm not 100% sure about the implications for other implementations.  So please understand this PR more like a bug report 🙂  In general, I saw that in https://github.com/LWJGL/lwjgl3/issues/875#issuecomment-1501178071 it is mentioned that you want to get rid of (ab)using `reserved3`, so I think that's the better fix going forward.

### DIFF
--- a/modules/lwjgl/core/src/main/java/org/lwjgl/system/ThreadLocalUtil.java
+++ b/modules/lwjgl/core/src/main/java/org/lwjgl/system/ThreadLocalUtil.java
@@ -186,7 +186,13 @@ public final class ThreadLocalUtil {
     public static void setFunctionMissingAddresses(int functionCount) {
         long ptr = JNI_NATIVE_INTERFACE + CAPABILITIES_OFFSET;
 
+        // aka. reserved3
+        //
+        // OpenJDK: NULL
+        // EspressoVM: NULL
+        // GraalVM Native Image: pointer to UnimplementedWithJNIEnvArgument function (see #875)
         long currentTable = memGetAddress(ptr);
+
         if (functionCount == 0) {
             if (currentTable != FUNCTION_MISSING_ABORT) {
                 FUNCTION_MISSING_ABORT_TABLE = NULL;
@@ -195,7 +201,7 @@ public final class ThreadLocalUtil {
             }
         } else {
             // OpenJDK: NULL
-            // EspressoVM: NULL
+            // EspressoVM: pointer to VM internal
             // GraalVM Native Image: pointer to UnimplementedWithJNIEnvArgument function (see #875)
             long RESERVED0_NULL = memGetAddress(JNI_NATIVE_INTERFACE);
 
--- a/modules/lwjgl/core/src/main/java/org/lwjgl/system/ThreadLocalUtil.java
+++ b/modules/lwjgl/core/src/main/java/org/lwjgl/system/ThreadLocalUtil.java
@@ -210,7 +210,7 @@ public final class ThreadLocalUtil {
             // GraalVM Native Image: pointer to UnimplementedWithJNIEnvArgument function (see #875)
             long RESERVED0_NULL = memGetAddress(JNI_NATIVE_INTERFACE);
 
-            boolean slotAvailable = currentTable == RESERVED3_NULL; // on HotSpot or Espresso
+            boolean slotAvailable = currentTable == NULL; // on HotSpot or Espresso
             slotAvailable |= currentTable == RESERVED0_NULL; // on Native Image
 
             if (!slotAvailable) {
--- a/modules/lwjgl/core/src/main/java/org/lwjgl/system/ThreadLocalUtil.java
+++ b/modules/lwjgl/core/src/main/java/org/lwjgl/system/ThreadLocalUtil.java
@@ -188,7 +188,7 @@ public final class ThreadLocalUtil {
 
         long currentTable = memGetAddress(ptr);
         if (functionCount == 0) {
-            if (currentTable != RESERVED3_NULL) {
+            if (currentTable != FUNCTION_MISSING_ABORT) {
                 FUNCTION_MISSING_ABORT_TABLE = NULL;
                 getAllocator().free(currentTable);
                 memPutAddress(ptr, NULL);
--- a/modules/lwjgl/core/src/main/java/org/lwjgl/system/ThreadLocalUtil.java
+++ b/modules/lwjgl/core/src/main/java/org/lwjgl/system/ThreadLocalUtil.java
@@ -93,11 +93,6 @@ public final class ThreadLocalUtil {
     /** The offset in JNIEnv at which to store the pointer to the capabilities array. */
     private static final int CAPABILITIES_OFFSET = 3 * POINTER_SIZE;
 
-    // OpenJDK: NULL
-    // EspressoVM: NULL
-    // GraalVM Native Image: pointer to UnimplementedWithJNIEnvArgument function (see #875)
-    private static long RESERVED3_NULL = memGetAddress(JNI_NATIVE_INTERFACE + CAPABILITIES_OFFSET);
-
     /** The number of pointers in the JNIEnv struct. */
     private static final int JNI_NATIVE_INTERFACE_FUNCTION_COUNT;
 
@@ -199,7 +194,15 @@ public final class ThreadLocalUtil {
                 memPutAddress(ptr, NULL);
             }
         } else {
-            if (currentTable != RESERVED3_NULL) {
+            // OpenJDK: NULL
+            // EspressoVM: NULL
+            // GraalVM Native Image: pointer to UnimplementedWithJNIEnvArgument function (see #875)
+            long RESERVED0_NULL = memGetAddress(JNI_NATIVE_INTERFACE);
+
+            boolean slotAvailable = currentTable == NULL; // on HotSpot or Espresso
+            slotAvailable |= currentTable == RESERVED0_NULL; // on Native Image
+
+            if (!slotAvailable) {
                 throw new IllegalStateException("setFunctionMissingAddresses has been called already");
             }
             if (currentTable != NULL) {
--- a/modules/lwjgl/core/src/main/java/org/lwjgl/system/ThreadLocalUtil.java
+++ b/modules/lwjgl/core/src/main/java/org/lwjgl/system/ThreadLocalUtil.java
@@ -90,9 +90,6 @@ public final class ThreadLocalUtil {
     /** The global JNIEnv. */
     private static final long JNI_NATIVE_INTERFACE = memGetAddress(getThreadJNIEnv());
 
-    /** The offset in JNIEnv at which to store the pointer to the capabilities array. */
-    private static final int CAPABILITIES_OFFSET = 3 * POINTER_SIZE;
-
     /** The number of pointers in the JNIEnv struct. */
     private static final int JNI_NATIVE_INTERFACE_FUNCTION_COUNT;
 
@@ -105,6 +102,9 @@ public final class ThreadLocalUtil {
      * <p>The array size depends on whether OpenGL or OpenGL ES is used.</p>
      */
     private static long FUNCTION_MISSING_ABORT_TABLE = NULL;
+
+    /** The offset in JNIEnv at which to store the pointer to the capabilities array. */
+    private static final int CAPABILITIES_OFFSET = 3 * POINTER_SIZE;
 
     static {
         int JNI_VERSION = GetVersion();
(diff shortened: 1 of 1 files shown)


modules/lwjgl/core/src/main/java/org/lwjgl/system/ThreadLocalUtil.java
```diff
@@ -184,21 +189,17 @@ public static void setCapabilities(long capabilities) {
...
-        // OpenJDK: NULL
-        // GraalVM Native Image: pointer to UnimplementedWithJNIEnvArgument function (see #875)
-        long RESERVED0_NULL = memGetAddress(JNI_NATIVE_INTERFACE);
-
         long ptr = JNI_NATIVE_INTERFACE + CAPABILITIES_OFFSET;
 
         long currentTable = memGetAddress(ptr);
         if (functionCount == 0) {
-            if (currentTable != RESERVED0_NULL) {
+            if (currentTable != RESERVED3_NULL) {
```
gilles-duboscq, line 196: What about checking for `currentTable == FUNCTION_MISSING_ABORT_TABLE` (i.e., "is it the value we set").
And resetting it to `RESERVED3_NULL` instead of `NULL` (i.e., reset it to "the value that was there before").

modules/lwjgl/core/src/main/java/org/lwjgl/system/ThreadLocalUtil.java
```diff
@@ -184,21 +189,17 @@ public static void setCapabilities(long capabilities) {
...
                 FUNCTION_MISSING_ABORT_TABLE = NULL;
                 getAllocator().free(currentTable);
                 memPutAddress(ptr, NULL);
             }
         } else {
-            if (currentTable != RESERVED0_NULL) {
+            if (currentTable != RESERVED3_NULL) {
```
gilles-duboscq, line 202: What is this check really trying to achieve? If it's about checking whether `reserved3` looks "available" then maybe a better heuristic is to assume that
* `NULL` means available (the common case)
* same value as `reserved0` likely means "available" for something  like GraalVM Native Image which fills `reserved0-3` with some non-null "unimplemented" value.

modules/lwjgl/core/src/main/java/org/lwjgl/system/ThreadLocalUtil.java
```diff
@@ -184,21 +184,25 @@ public static void setCapabilities(long capabilities) {
...
-            if (currentTable != RESERVED0_NULL) {
+            if (currentTable != FUNCTION_MISSING_ABORT) {
                 FUNCTION_MISSING_ABORT_TABLE = NULL;
                 getAllocator().free(currentTable);
                 memPutAddress(ptr, NULL);
             }
         } else {
-            if (currentTable != RESERVED0_NULL) {
+            // OpenJDK: NULL
+            // EspressoVM: NULL
```
gilles-duboscq, line 198: `reserved0` can be some non-null value on espresso

modules/lwjgl/core/src/main/java/org/lwjgl/system/ThreadLocalUtil.java
```diff
@@ -184,21 +193,27 @@ public static void setCapabilities(long capabilities) {
...
+                memPutAddress(ptr, RESERVED3_NULL);
             }
         } else {
-            if (currentTable != RESERVED0_NULL) {
+            // OpenJDK: NULL
+            // EspressoVM: pointer to VM internal
+            // GraalVM Native Image: pointer to UnimplementedWithJNIEnvArgument function (see #875)
+            long RESERVED0_NULL = memGetAddress(JNI_NATIVE_INTERFACE);
+
+            boolean slotAvailable = currentTable == RESERVED3_NULL; // on HotSpot or Espresso
```
gilles-duboscq, line 213: I think it's better to check against `NULL` here, that allows detecting VMs or environment which put something meaningful in `reserved3`.

xxDark:
I have ran into this issue myself, what if all reversed slots are in use? Won't it be better to copy JNIEnv with one more pointer in the function table? `nsetupEnvData` copies the JNI function table already.

Spasi:
Hey @lewurm and @gilles-duboscq,

I'm not convinced this is necessary. It looks like EspressoVM uses the reserved fields of the `JNIInvokeInterface` struct (i.e. `JavaVM`) and not those of the `JNINativeInterface` struct (i.e. `JNIEnv`) which LWJGL (ab)uses.

Is the current implementation actually broken on EspressoVM? Could the issue be related to something else?

Spasi:
I've now done some testing with EspressoVM and it indeed populates `reserved0-3`, just not with the values mentioned above. This is my current understanding of the situation in each VM:

```
VM        | HotSpot | GraalVM Native Image (#875)     | EspressoVM / Truffle (#1003)     |
----------+---------+---------------------------------+----------------------------------+
reserved0 | NULL    | UnimplementedWithJNIEnvArgument | struct NespressoEnv *            |
reserved1 | NULL    | UnimplementedWithJNIEnvArgument | struct MokapotNativeInterface_ * |
reserved2 | NULL    | UnimplementedWithJNIEnvArgument | unset_function_error             |
reserved3 | NULL    | UnimplementedWithJNIEnvArgument | unset_function_error             |
```

Based on the above, the following implementation seems to work:

```java
private static final int CAPABILITIES_OFFSET = 3 * POINTER_SIZE;

private static final long RESERVED_NULL = memGetAddress(JNI_NATIVE_INTERFACE + CAPABILITIES_OFFSET);

public static void setFunctionMissingAddresses(int functionCount) {
    long ptr = JNI_NATIVE_INTERFACE + CAPABILITIES_OFFSET;

    long currentTable = memGetAddress(ptr);
    if (functionCount == 0) {
        if (currentTable == FUNCTION_MISSING_ABORT_TABLE && FUNCTION_MISSING_ABORT_TABLE != NULL) {
            FUNCTION_MISSING_ABORT_TABLE = NULL;
            getAllocator().free(currentTable);
            memPutAddress(ptr, RESERVED_NULL);
        }
    } else {
        if (currentTable != RESERVED_NULL) {
            throw new IllegalStateException("setFunctionMissingAddresses has been called already");
        }
        if (currentTable != NULL) {
            // check reserved0 to see if this Native Image or EspressoVM. EspressoVM will not have the reserved NULL here.
            if (memGetAddress(JNI_NATIVE_INTERFACE) == RESERVED_NULL) {
                // silently abort on Native Image, the global JNIEnv object lives in read-only memory by default. (see #875)
                return;
            }
        }

        FUNCTION_MISSING_ABORT_TABLE = getAllocator().malloc(Integer.toUnsignedLong(functionCount) * POINTER_SIZE);
        for (int i = 0; i < functionCount; i++) {
            memPutAddress(FUNCTION_MISSING_ABORT_TABLE + Integer.toUnsignedLong(i) * POINTER_SIZE, FUNCTION_MISSING_ABORT);
        }

        memPutAddress(ptr, FUNCTION_MISSING_ABORT_TABLE);
    }
}
```

I can successfully run LWJGL applications with EspressoVM (with `-truffle`) on macOS. However, I'm having trouble on Windows which is unrelated to `ThreadLocalUtil`. It crashes much earlier, without any output, when calling any native method of the `WinBase` class. Output with `--log.java.level=FINEST` looks like:

```
[...]
[java::VM] FINEST: JVM_FindLibraryEntry(LibFFILibrary(140731941978112), Java_org_lwjgl_system_jni_JNINativeInterface_nNewDirectByteBuffer) -> NativePointer(140731942113088)
[java] FINEST: Initializing: java/lang/invoke/LambdaForm$DMH+1069
[java] FINEST: Initializing: java/lang/invoke/LambdaForm$DMH+1070
[java] FINEST: Initializing: org/lwjgl/system/MemoryUtil$$Lambda+1072
[java] FINEST: Initializing: java/util/function/LongPredicate
[java] FINEST: Initializing: java/lang/invoke/LambdaForm$DMH+1073
[java] FINEST: Initializing: org/lwjgl/system/Pointer
[java] FINEST: Initializing: org/lwjgl/system/MemoryAccessJNI
[java::VM] FINEST: JVM_FindLibraryEntry(LibFFILibrary(140731941978112), Java_org_lwjgl_system_MemoryAccessJNI_malloc) -> NativePointer(140731942110592)
[java::VM] FINEST: JVM_FindLibraryEntry(LibFFILibrary(140731941978112), Java_org_lwjgl_system_MemoryAccessJNI_calloc) -> NativePointer(140731942110608)
[java::VM] FINEST: JVM_FindLibraryEntry(LibFFILibrary(140731941978112), Java_org_lwjgl_system_MemoryAccessJNI_realloc) -> NativePointer(140731942110624)
[java::VM] FINEST: JVM_FindLibraryEntry(LibFFILibrary(140731941978112), Java_org_lwjgl_system_MemoryAccessJNI_free) -> NativePointer(140731942110640)
[java::VM] FINEST: JVM_FindLibraryEntry(LibFFILibrary(140731941978112), Java_org_lwjgl_system_MemoryAccessJNI_aligned_1alloc) -> NativePointer(140731942110656)
[java::VM] FINEST: JVM_FindLibraryEntry(LibFFILibrary(140731941978112), Java_org_lwjgl_system_MemoryAccessJNI_aligned_1free) -> NativePointer(140731942110688)
[java::VM] FINEST: JVM_FindLibraryEntry(LibFFILibrary(140731941978112), Java_org_lwjgl_system_MemoryAccessJNI_getPointerSize) -> NativePointer(140731942110576)
[java] FINEST: Initializing: java/lang/invoke/LambdaForm$DMH+1075
[java] FINEST: Initializing: java/lang/invoke/LambdaForm$DMH+1076
[java] FINEST: Initializing: org/lwjgl/system/MemoryUtil$$Lambda+1077
[java] FINEST: Initializing: java/lang/invoke/LambdaForm$DMH+1078
[java] FINEST: Initializing: java/lang/invoke/LambdaForm$MH+1079
[java] FINEST: Initializing: java/lang/invoke/LambdaForm$DMH+1080
[java] FINEST: Initializing: java/lang/invoke/LambdaForm$DMH+1081
[java] FINEST: Initializing: org/lwjgl/system/MemoryUtil$$Lambda+1082
[java] FINEST: Initializing: org/lwjgl/system/FunctionProvider
[java] FINEST: Initializing: org/lwjgl/system/NativeResource
[java] FINEST: Initializing: org/lwjgl/system/MemoryStack
[java] FINEST: Initializing: org/lwjgl/system/MemoryStack$$Lambda+1087
[java] FINEST: Initializing: org/lwjgl/BufferUtils
[java] FINEST: Initializing: org/lwjgl/system/windows/WinBase
[java::VM] FINEST: JVM_FindLibraryEntry(LibFFILibrary(140731941978112), Java_org_lwjgl_system_windows_WinBase_nGetModuleHandle) -> NativePointer(140731942118960)
```

The process simply dies after the last line (and before the native method returns). I have no idea what might be causing this.

Spasi:
After more fixes and testing, it turns out that abusing `JNIEnv *` for thread-local storage is fundamentally incompatible with the JNI implementation in EspressoVM:

- On Hotspot, each thread uses a different `JNIEnv *`, pointing to the corresponding field of its `JavaThread` instance. This is what allows LWJGL to inject `JNIEnv` copies and use the `reserved` fields for storage. Also, all threads normally point to the same, global `JNIEnv` struct, which is never deallocated. This simplifies the cleanup LWJGL has to do on thread exit.

- On EspressoVM, JNI calls use a single, global `JNIEnv *`, allocated dynamically and reused by all threads. Therefore, abusing it for thread-local storage is not possible. Cleanup is also complicated because the `JNIEnv` it points to is freed on VM exit.

As a partial fix, I've replaced the thread-local storage of `errno/GetLastError()` with virtual output parameters, (optionally) passed explicitly by the caller. This approach is similar to `Linker.Option.CaptureCallState` in Project Panama's FFM API. With this change, LWJGL no longer uses `JNIEnv::reserved2` and there is no need for `JNIEnv` copy injections outside of OpenGL(ES).

https://github.com/LWJGL/lwjgl3/commit/2b12f5e9fb52b1f81482d4f2aee39894bfd7a3f9

After the above commit, LWJGL tests pass and demos that do not use OpenGL work correctly on EspressoVM. It is a breaking change, but it should affect very few users, that make system calls under `org.lwjgl.system.linux.*`,  `org.lwjgl.system.windows.*` and `org.lwjgl.opengl.WGL`.

lewurm:
> * On EspressoVM, JNI calls use a single, global `JNIEnv *`, allocated dynamically and reused by all threads. Therefore, abusing it for thread-local storage is not possible.

ahhh 😅 


Thanks for the investigation and for the fix @Spasi!

Issue #1008, opened 2024-10-11
Migrate to JSpecify nullability annotations


TheMrMilchmann:
### Description

Back in #344, we introduced nullability marker annotations to the code base. I had two goals in mind when working on this change:

1. Explicit nullability information for LWJGL (especially binding methods),
2. Enhance Kotlin compatibility by avoiding [platform types](https://kotlinlang.org/docs/java-interop.html#null-safety-and-platform-types).

We made good progress towards both of these goals. However, some caveats still apply to this day:

1. Although the nullability information is available, it is not enforced without the use of third-party tools. Java (or the Java compiler) itself does not use any semantic information from these annotations.
2. Strict support for JSR305 annotations has to be enabled explicitly by passing the `-Xjsr305=strict` flag to the Kotlin compiler.

By adding the required markers, we effectively provided the requirements for null analysis. Setting up the checks was still on the user's site. Although that was unfortunate, the lack of nullability support in the languages (for both, Java and Kotlin users) held us back.

At the time, using an unofficial implementation of the dormant JSR305 implementation seemed like a good idea: Since it was the closest there was to an official specification, it was supported by most tooling and had widespread use. However, there are three issues with this implementation:

1. Although the JSR305 annotations are supported by a broad set of tools, the _specification_ is complex and not standardized. Thus, many different solutions with the same goals were developed.
2. The JSR305 annotations sit in the `javax.annotations` package and were developed before the JPMS imposed restrictions on shared packages. Thus, using JSR305 annotations alongside some JavaEE modules could cause issues on certain versions.
3. The use of the `javax` package raises licensing questions and concerns. Specifically, Oracle's Java licensing forbids use of this package for unofficial code. More on this can be found in the [Guava issue for replacing JSR305 annotations](https://github.com/google/guava/issues/2960).

[JSpecify](https://jspecify.dev/) is a joint effort to (finally) standardize nullability markers. Many companies are involved - including Google, Oracle, and JetBrains. Consequentially, preliminary JSpecify support has already been available in IDEA for a while. Similarly, Kotlin supported JSpecify for a while already, and 2.0.20 fixed (most of) the remaining open issues. [There are plans](https://youtrack.jetbrains.com/issue/KT-55586/Handle-nullability-from-jspecify-annotations-properly) to enable strict handling of JSpecify annotations by default in 2.1.0.

[...]

PR #1019, opened 2024-11-13
feat: migrate to JSpecify

TheMrMilchmann:
See #1008 for a detailed description.

Key changes:

- All occurrences of `javax.annotation.Nullable` have been replaced with `org.jspecify.annotations.Nullable`
  - This introduced a conceptual change as nullability effectively becomes a use-site attribute of the type. Where a method previously returned "String or null", it now returns a "nullable String". To reflect this, the location of the marker annotation was (an in a few cases had to be) changed.
  - `@Nullable public static Something.Buffer ...` becomes `public static Something.@Nullable Buffer ...`
  - The `@Nullable` marker now directly precedes the type before the `@NativeType` annotation where applicable.
- Likewise, `org.lwjgl.system.NonnullDefault` has been removed and replaced with `org.jspecify.annotations.NullMarked`.

This is a draft PR for the following reasons:

1. We need to decide whether we want to expose JSpecify as an API dependency. JSpecify recommends this, but using them as compile-only dependencies (just like we currently do with JSR305 annotations) would also work fine.
2. I'm not fully happy with the changes to the generator. If you have a better idea, please let me know.


modules/lwjgl/core/src/main/java/org/lwjgl/system/SharedLibraryLoader.java
```diff
@@ -36,9 +35,7 @@ final class SharedLibraryLoader {
 
     private static final Lock EXTRACT_PATH_LOCK = new ReentrantLock();
 
-    @GuardedBy("EXTRACT_PATH_LOCK")
```
TheMrMilchmann, line 39: This is the only use of JSR305 annotations that can't be replaced with JSpecify. We _could_ keep JSR305 just for that but I figured it's not worth it.

Spasi:
Great job @TheMrMilchmann, thanks!

TheMrMilchmann:
> Is it valid syntax to have a type, the annotation, and then .Buffer?
> 
> https://github.com/LWJGL/lwjgl3/blob/2607557077ce33a35521f1fa10a79dfea050d440/modules/lwjgl/assimp/src/generated/java/org/lwjgl/assimp/AIBone.java#L235

@SWinxy This is not just valid, but the only correct syntax for the `TYPE_USE` syntax. In simple terms, these annotations must appear next to the concert type name (or `[]`) they apply to. `AIBone` is just part of the qualifier for its nested `Buffer`. This is specified in [JLS 9.7.4](https://docs.oracle.com/javase/specs/jls/se8/html/jls-9.html#jls-9.7.4).

Some examples from [JSpecify's docs](https://jspecify.dev/docs/user-guide/#type-use-annotation-syntax) (which apply to all `TYPE_USE` annotations):

> 1. For a nested static type like Map.Entry, if you want to say that the value can be `null` then the syntax is `Map.@Nullable Entry`. You can often avoid dealing with this by importing the nested type directly, but in this case `import java.util.Map.Entry` might be undesirable because `Entry` is such a common type name.
>
> 2. For an array type, if you want to say that the elements of the array can be `null` then the syntax is `@Nullable String[]`. If you want to say that the array itself can be `null` then the syntax is `String @Nullable []`. And if both the elements and the array itself can be `null`, the syntax is `@Nullable String @Nullable []`.

SWinxy:
Wild! Thanks for the info.

Issue #174, opened 2016-03-29
SDL2 Binding?


renannprado:
What do you guys think?

huhlig:
What value are you looking to derive from SDL2 you cannot get from GLFW?

renannprado:
It's been a long time since I last used it, BUT when I did more or less a year ago, the big win for me was the simplified graphics API over OpenGL one.
But I'm certainly not expert in OpenGL nor SDL2.
Other people will probably have better arguments (against or for) than me.

huhlig:
So you want to use SDL2 instead of GLFW inside a Java OpenGL library to draw graphics without OpenGL?

RegaliaSly:
While it would be nice to see SDL2 get a working port to Java, it would have no point in LWJGL as GLFW already handles everything in terms of window, input and context handling.

dustContributor:
Pretty much, want something like SDL2? Use libGDX.

XenoAmess:
glfw have no method like SDL_JoystickRumble in SDL.///
But you can use Jamepad instead so that is no problem.

PR #1033, opened 2024-12-28
feat: SDL 3.1.8 bindings

eigenraven:
This PR adds bindings for [SDL 3.1.8](https://libsdl.org/), which provides an alternative featureful library for window management, graphics, audio and other platform abstractions for video games.

Notable omissions from the bindings are:
- Various string handling and math functions from `SDL_stdinc.h` that have Java equivalents in the standard library
- Process spawning and environment variable control - this is better delegated to the JVM
- Thread spawning and atomics - these require JVM cooperation, so exposing them directly via FFI would be inefficient and/or unsafe
- Shared library loading - LWJGL core already provides a much better alternative to this in the form of libffi

I have ported the GLFW Gears.java demo to SDL to show how the API can be used to manage a GL context and handle keyboard input.

image

Original PR message:

After the discussion on discord where interest for SDL3 bindings was expressed, I thought I'd start on them to try to help. Currently I've just set up the first header of the library that holds some utility functions and tested it locally on my Linux install. 

If possible, I'd like to get some feedback if I'm on the right path and if I should make any changes to the code structure before I run the extractor and adjust the bindings on the 70 or so SDL headers (probably fewer, as quite a lot of them are wrappers for functionality Java already provides - like threads, atomics and synchronization primitives)

Related to https://github.com/LWJGL/lwjgl3/issues/174


Spasi:
Hey @eigenraven, thanks for getting this started!

It's looking good structure-wise, except you don't need to add the SDL headers to LWJGL. We'll build the SDL library separately, in an [LWJGL-CI](https://github.com/LWJGL-CI/) fork, the LWJGL build itself won't have to do any native compilation. Except in the rare case of having to support a function that passes/returns a struct by-value. Even then, we may choose to support such functions dynamically with libffi, if they're only one or two. Anyway, you'll know you may need the headers if you see `.c` files being generated under `modules/lwjgl/sdl3/src/generated/c/`.

For reference, other bindings that should have similar structure to SDL are: GLFW, OpenAL Soft, bgfx, freetype, etc. (you'll find all of them in LWJGL-CI).

You should probably wait a bit before doing more work on this though. The next LWJGL version (`3.4.0`) is going to remove documentation from bindings. I noticed that you've already added a lot of documentation (in `SDL_stdinc.kt`) and all that is going away. The Kotlin DSL is also going to be simplified. I'm almost done with this refactoring and will be pushing it in a few days. Better wait to rebase on top of that before adding more headers.

Spasi:
Hey @eigenraven,

Current HEAD now has the LWJGL 3.4.0 changes (removed binding documentation and simplified generator DSL). I rebased this PR and made a few changes so that you don't waste time with the boring stuff:

- Removed SDL headers.
- Renamed `sdl3` module/package to just `sdl`.
- Updated the Kotlin templates with the new generator API without documentation.

It should be nice and clean now and you may continue adding more SDL bindings.

I do have another concern though. SDL's API contains a lot of functionality that is already available in the JDK, in just as good or better form. One approach is to make this simple, expose everything and let LWJGL users decide what to use or not. Another approach would be to filter out such functionality. Examples:

- `SDL_atomic.h`
- `SDL_filesystem.h`
- `SDL_iostream.h`
- `SDL_stdinc.h`: math functions, most of the string functions, etc.
- `SDL_process.h`
- `SDL_thread.h`
- `SDL_time.h`

And so on. I haven't gone through everything, but I'm sure there is a ton of functionality that would just be unnecessary noise to a Java developer.

eigenraven:
Thanks for doing the rebase! Regarding filtering the available functions, I definitely think a lot of the math, atomics, string, quicksort etc. functions should just be skipped (the math functions that are in right now were mostly included so I could run a test using my system native library before porting the other headers). I lean towards leaving the rest in unless it's fundamentally unsafe (like threading bypassing Java's own threading system) or complicated to bind to.

The filesystem APIs might actually be useful, because they expose os-specific recommended local storage for apps, including e.g. Steam Cloud integration for games published on Steam. There are also plans to implement async io in SDL using io_uring on Linux, another unique feature.
Spawning Java threads from C would probably be really unsafe and I'll skip that, but leave in the TLS management that's needed to run cleanup on SDL's own TLS storage used by some of its subsystems when used from multiple threads.
SDL timers might be useful as a different way of accessing platform timers, because what Java uses and what SDL uses might not be the same everywhere, they're a really simple API so I think it's best to just leave in.

The message box and file dialog API should also provide a nice alternative to tinyfd/nfd, especially as there's discussions about removing tinyfd.

I'll start on porting more headers now.,

One remaining thing to consider is AWT conflicts on macOS, it will probably require -XstartOnFirstThread just like GLFW unless we implement something like glfw_async for SDL (maybe the upstream could be convinced to include the necessary mechanisms? I'm not a macOS developer at all so I don't know what glfw_async actually does under the hood)

Spasi:
Requiring `-XstartOnFirstThread` will be fine (and expected).

Definitely skip the async treatment for now, it's complicated and a pain to maintain. Manageable for GLFW because it's been very stable, but SDL moves too quickly. I also don't think they'd want to waste time with something like that upstream.

eigenraven:
I just saw in the 3.1.8 changelog that they did add https://wiki.libsdl.org/SDL3/SDL_IsMainThread and https://wiki.libsdl.org/SDL3/SDL_RunOnMainThread for better supporting their own SDL_main event loop, so there is a chance :thinking: I'll leave it be for now though

eigenraven:
@Spasi the bindings are now complete and ready to review including a glfw Gears.java demo port. The only thing I think I'll have to leave to you is the CI setup for natives. Let me know if there's anything else I should do.
For Windows SDL publishes pre-built DLLs in their github releases that should be perfectly fine to use (and reduce chance of antivirus issues), but they also provide extensive build documentation for all platforms in their `docs/` directory in the source tree, including required package lists for Linux etc.
The formatting in some files might not be entirely consistent, I tried to follow what other files did but I didn't see an autoformatter config so I can't guarantee it's correct everywhere.

Spasi:
Thank you @eigenraven!

renannprado:
@eigenraven almost 10 years later, a hero appears. That's all I wanted to say! Thanks for the contribution and thanks @Spasi for keeping it up as well!

PR #1037, opened 2025-01-12
fix(extract): run Swing code on the correct event dispatcher thread

eigenraven:
Swing requires running all setup and widget interaction code on the [AWT dispatcher thread](https://docs.oracle.com/javase/tutorial/uiswing/concurrency/dispatch.html), so in this PR I moved the app initialization over to that thread.

I've had some random native crashes when running the extractor locally, this fix might or might not have fixed them (they're too unpredictable to be sure), but the hs_err stacktraces looked to come from a variety of native AWT code failing to read standard java data structures which could be due to some state corruption due to a data race between threads.


Spasi:
@eigenraven Thanks!

PR #1040, opened 2025-01-27
Update tinyfiledialogs to v3.19.1

JakobDev:
This version fixes a bug, that when running on Linux under Wayland but with XWayland disabled, tinyfiledialog thinks it not running in a graphical session.


Spasi:
Thanks @JakobDev!

Issue #1031, opened 2024-12-21
no window rendered running HelloOpenXRGL sample


farmboy0:
### Version

3.3.5

### Platform

Linux x64

### JDK

GraalVM 17

### Module

GLFW, OpenXR, OpenGL

### Bug description

I tried running the HelloOpenXRGL on linux with LWJGL 3.3.5 but there are 2 problems with it.
1. you need to add GL21.GL_SRGB8 to the desiredSwapchainFormats at line 600-603
2. The real problem I see though is that the 2D window the sample opens is empty on X11 and not even showing the window border on wayland.

Im running Gentoo Linux with GLFW 3.4

knokko:
It has been a long time since I wrote this sample, and I don't really use OpenGL nowadays. Could you elaborate point (1)? *Why* is it a good idea to add this desired swapchain format?

Regarding point (2), that is desired! I never wanted to create a window; I just want to render on the headset. But... creating an OpenGL context without creating a window is surprisingly hard, so the easiest way to do this, is by creating a small invisible window.

farmboy0:
The new format is one of the supported ones with SteamVr on linux, the others in the list arent.

knokko:
Do you know the full list of formats supported by SteamVR?

farmboy0:
There are 5:
GL_SRGB8	35905	0x8c41
GL_SRGB8_ALPHA8	35907	0x8c43
GL_DEPTH_COMPONENT16	33189	0x81a5
GL_DEPTH_COMPONENT24	33190	0x81a6
GL_DEPTH_COMPONENT32F	36012	0x8cac

farmboy0:
As for the second point: No its not desired. If a window is programmed to show and render into it better be. Its not even desired to be able to create VR apps that have no window at all. There would be no way for the OS to determine the current window to send input events to.

knokko:
> There are 5: GL_SRGB8 35905 0x8c41 GL_SRGB8_ALPHA8 35907 0x8c43 GL_DEPTH_COMPONENT16 33189 0x81a5 GL_DEPTH_COMPONENT24 33190 0x81a6 GL_DEPTH_COMPONENT32F 36012 0x8cac

Ok, then I think that indeed `GL_SRGB8` would be the best choice. Can you create a pull request for this? (I don't have the right dev environment at the moment, so it would be nice if you could do it.)



[...]

PR #1060, opened 2025-04-29
fix(demo): add GL_SRGB8 to swapchain formats of HelloOpenXRGL

knokko:
According to https://github.com/LWJGL/lwjgl3/issues/1031, this GL format is one of the few OpenXR swapchain formats supported by SteamVR, so it is good to add it to the desired swapchain formats.


Spasi:
Thanks!

PR #1063, opened 2025-05-15
Vulkan HelloTriangle 1.3 port

controllerface:
Adds a port of the Vulkan HelloTriangle found in the Vulkan-Samples repo here:

https://github.com/KhronosGroup/Vulkan-Samples/tree/main/samples/api/hello_triangle_1_3

The version uses the glfw bindings.

### DIFF
--- a/modules/samples/src/test/java/org/lwjgl/demo/vulkan/khronos/HelloTriangle_1_3.java
+++ b/modules/samples/src/test/java/org/lwjgl/demo/vulkan/khronos/HelloTriangle_1_3.java
@@ -1,0 +1,1279 @@
+/*
+ * Copyright LWJGL. All rights reserved.
+ * License terms: https://www.lwjgl.org/license
+ */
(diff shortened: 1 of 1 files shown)


controllerface:
I updated my graphics drivers yesterday and figured I would try to run this demo again, just in case, and it no longer runs. So going to try and debug this. If I can't figure it out relatively quickly I will withdraw this PR.

controllerface:
Ok, it's working again. Thanks to TheNEET2 from the LWJGL discord for giving the code a review pass and helping me find the remaining issues.

I also made sure to test this on more than just my AMD desktop, it also works on my Nvidia/Intel Laptop. There was also one minor validation error when resizing the window that I addressed with a comment describing what led to the change. Instead of recycling a semaphore when resizing, it is destroyed allowing a new one to be created automatically.

Spasi:
Thank you @controllerface!

PR #1086, opened 2025-10-07
add shell.nix

lego-eden:
This adds a `shell.nix` file which makes i easy to get going with developing on nix. I believe anyone who wants to use nix to develop would make their own shell.nix file, and this is simply my attempt at writing one. It may not be perfect, but it works. Simply run `nix-shell` in the project root, and you get all the tools and libraries you need for developing LWJGL. If there is a need to add a new dependency.

EDIT: I just found out about the commit message conventions. I think my commit msg is wrong, but do not really know what prefix it should have. I update the message and force push the change if I need to, I just need to know what it should be updated to.

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -13,10 +13,6 @@ let
     xorg.libXxf86vm
     wayland
     libxkbcommon
-    gtk3
-    gtk3-x11
-    gdk-pixbuf
-    dbus
   ];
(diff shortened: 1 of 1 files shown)


lego-eden:
This needed more work, I think I will rework this at some point. Will keep the branch around if anyone wants to have a look

PR #1091, opened 2025-11-09
feat(renderdoc) add RenderDoc bindings

io7m:
This implements bindings for the RenderDoc debugger. The bindings must be explicitly initialized before use via the `RenderDoc.create()` method, and exceptions must be caught; if the RenderDoc debugger is not running, the RenderDoc native library will not have been injected into the application's process and the initialization will fail. A sample is included to demonstrate how to do this correctly.

See: https://renderdoc.org
See: https://renderdoc.org/docs/in_application_api.html


io7m:
This _seems_ correct to me, but it is slightly awkward to test as you need to run `ant` to run the included sample under `qrenderdoc`. `qrenderdoc` must be told to trace all child processes. I've tried it a bit with a test application and didn't see any problems, and the library seems to correctly tell you when you're not using `qrenderdoc`.

PR #1095, opened 2025-12-01
feat(sdl) Provide main callbacks as an alternate way of structuring SDL apps

eigenraven:
The docs suggested it was not usable from other languages, but https://github.com/libsdl-org/SDL/issues/14513 indicates a valid way of using the main callbacks across a FFI boundary.

The callbacks don't do much on Windows/Linux/macOS (they are required on Android/iOS/Xbox/other consoles for system setup), mostly they handle calling the iterate function while the window is resized on Windows - but they might become more important in the future, and are the recommended way of using SDL.

I've also updated and tested the Gears demo with the callback structure as an option.

The core library could no longer build on my Arch Linux PC because unused parameters warnings became enabled and are not part of Wpedantic, I've added it to the suppressions in LinuxConfig.h


Spasi:
Thank you @eigenraven!

PR #1132, opened 2026-06-23
build(core) harden GitHub Actions workflows and add zizmor scanning

httpdigest:
Mostly hardening actions via pinned sha digest to avoid GitHub actions supply-chain attacks via compromised GitHub actions by tag.
Also add a dependabot configuration to keep GitHub actions in their latest versions.


Spasi:
Thanks @httpdigest!

